PR #6, opened 2024-03-18
Edit Payment Request form

alfonsograziano:


apps/payments/app/[locale]/(hosted)/paymentSetup/edit/[request_id]/page.tsx
```diff
@@ -0,0 +1,269 @@
...
+    // Here it stops for whatever reason...
+    console.log(result);
+
+    redirect(`/paymentSetup/requests/${paymentRequestId}`);
+  } catch (e) {
+    await client.query("ROLLBACK");
+    console.log(e);
+  }
+
+  return;
```
Comment (peschina) [inline apps/payments/app/[locale]/(hosted)/paymentSetup/edit/[request_id]/page.tsx:80]: Looks like `redirect()` throws an error that Next then handles to stop code execution, so we need to move it outside of the try catch block. I think we can also remove `user_id` from the query, we just need `payment_request_id`

```suggestion
    await client.query(
      `update payment_requests 
    set title = $1, description = $2, reference = $3, amount = $4, redirect_url = $5, allow_amount_override = $6 
    where payment_request_id = $7`,
      [
        data.title,
        data.description,
        data.reference,
        data.amount,
        data.redirectUrl,
        data.allowAmountOverride,
        paymentRequestId,
      ],
    );
    await client.query("COMMIT");
  } catch (e) {
    await client.query("ROLLBACK");
    throw e;
  }
  redirect(`/paymentSetup/requests/${paymentRequestId}`);
```

Comment (alfonsograziano) [reply apps/payments/app/[locale]/(hosted)/paymentSetup/edit/[request_id]/page.tsx:80]: Good catch! At this step we can remove the user_id but when we create the APIs we need to keep that in mind. An user must be able to update only its requests

alfonsograziano:
Closing since there is some trash in the PR I'm not able to fix

PR #18, opened 2024-03-19
Fix tl payment flow

peschina:
- Fix stripe import
- Fix account holder name for Truelayer provider seeded (parenthesis are not allowed)


alfonsograziano (approved):
LGTM

PR #20, opened 2024-03-20
Create providers using APIs + autogenerated client

alfonsograziano:


alfonsograziano:
I looked into using `typescript-fetch` instead of `typescript-axios`, but it looks like there are some really annoying [bugs](https://github.com/OpenAPITools/openapi-generator/issues?q=is%3Aissue+is%3Aopen+typescript-fetch+) for now with this generator.

I suggest we keep Axios for now, and if things change, we can revisit our choice. FYI @peschina

apps/payments-backend/index.ts
```diff
@@ -1,26 +1,10 @@
...
-    },
-  },
-});
-
-app.register(fastifySwaggerUi, {
-  routePrefix: "/docs",
-});
-
-app.listen({ port: 8080 }, (err, address) => {
+app.listen({ port: 8001 }, (err, address) => {
```
alfonsograziano, line 7: We need to load the port from env variable

apps/payments-backend/plugins/auth.ts
```diff
@@ -0,0 +1,32 @@
...
+}
+
+declare module "fastify" {
+  interface FastifyRequest {
+    user?: User;
+  }
+}
+
+// Very secure, isn't it?
+const authPlugin: FastifyPluginAsync = async (fastify) => {
```
alfonsograziano, line 16: Temporary solution, will be revisited as soon as we have proper login in place

apps/payments/client/index.ts
```diff
@@ -0,0 +1,20 @@
+import axios from "axios";
+import { ProvidersApi } from "./autogenerated";
+
+const buildAxiosInstance = (userId) => {
+  const instance = axios.create({});
+  instance.defaults.headers.common["x-user-id"] = userId;
+  return instance;
+};
+
+const backendUrl = "http://localhost:8001";
```
alfonsograziano, line 10: We need to load this from env variable, `NEXT_PUBLIC_PAYMENTS_BACKEND_URL`

SamSalvatico (approved):
LGTM

PR #21, opened 2024-03-20
Fix Stripe transaction amount reported 

peschina:
Fix query to update transaction amount when using stripe


alfonsograziano (approved):
LGTM

PR #30, opened 2024-03-21
Docs improvements + guidelines docs

alfonsograziano:


README.md
```diff
@@ -1,50 +1,6 @@
 # Life Events Platform
 
-Life Events Platform, using the [OGCIO](https://www.ogcio.gov.ie/) building blocks.
+Welcome! ✨ In order to read this project's documentation, you need to just run a command: `npm i && npm start:docs`
```
GiuliaTeggi, line 3: Personally I liked to have the main setup instructions available in the main README, rather than having to start another service just to read them..but happy to hear other people's thoughts :)

documentation/docs/processes.md
```diff
@@ -0,0 +1,88 @@
...
+sidebar_label: Processes
+---
+
+# Processes
+
+In this document, you will find everything about how we work, the process we follow and the guidelines.
+
+## Git and Github
+
+Here's a draft for your software development project's Git and GitHub processes documentation in Markdown format. This template includes sections for branching, commit messages, and review processes. Feel free to customize it according to your project's specific guidelines and practices.
```
GiuliaTeggi, line 12: I feel like this paragraph could be removed?

README.md
peschina, line 3: Since this is an open source project I think we should leave at least a short description in the readme and a table of contents pointing to the different markdown files that we have in the docs. This way the documentation will be more easily available to those who don't wont to clone the project, install package and so on

documentation/docs/processes.md
```diff
@@ -0,0 +1,88 @@
...
+#### Creating Pull Requests
+
+1. Before creating a pull request, ensure your branch is up to date with `dev`.
+2. Create a pull request with a clear title and description. If applicable, you should stick to the PR template.
+3. Assign the pull request to yourself and tag relevant reviewers.
+
+#### Review Process
+
+1. Reviewers should provide constructive feedback and approve changes only when they meet project standards.
+2. Address feedback and make necessary adjustments. Re-request reviews if needed.
```
peschina, line 78: Do we want to add something regarding resolving conversations?

peschina (commented):
A few comments :)

README.md
alfonsograziano: Makes sense! Just to keep everything on the same place, what if we just link to the setup markdown file from here? 
In the end, if you don't want to start the Docs service, you can just read the docs in the `documentation/docs` folder in Markdown

documentation/docs/processes.md
alfonsograziano: Good catch :)

documentation/docs/processes.md
alfonsograziano: Good idea. I added the point. Feel free to suggest changes to the paragraph in case you want

peschina: I think the conversation should be resolved by the one that is opening the conversation. Ideally I image the flow to be:
- conversation is open by A
- comment is address by B
- B re-requests review from A
- A reviews, closes conversation and approves the PR

If the PR is already approved and the comment is not blocking the conversation can be resolved by the one that opened the PR.

Would like to hear what the others think though

GiuliaTeggi: I agree with @peschina, although this should only be a recommendation, as there are cases where it just makes sense by user A to resolve it (especially if the conversation is about a small fix) - otherwise it could get quite time consuming for everyone.

peschina: Yes, my thoughts is that this should be a guideline, then good sense always applies

alfonsograziano: addressed :)

README.md
alfonsograziano: Morning! Just addressed, please take a look when you have time :)

peschina (approved):
LGTM

PR #31, opened 2024-03-22
Delete payment request

alfonsograziano:
- Add "Delete" button on the payment request
- A payment request can be deleted only if there are no transactions yet
- An use case can be experimenting with the system, creating a Draft payment request or just creating it by mistake
- If there are already payment requests, the Payment request can only be "Disabled" in order to not accept new transactions (will be implemented in a new PR)

### DIFF
--- a/apps/payments/app/[locale]/(hosted)/paymentSetup/requests/[requestId]/page.tsx
+++ b/apps/payments/app/[locale]/(hosted)/paymentSetup/requests/[requestId]/page.tsx
@@ -28,6 +28,30 @@ async function createTransaction(requestId: string, formData: FormData) {
   redirect(requestId);
 }
 
+async function deletePaymentRequest(requestId: string) {
+  "use server";
+
+  await pgpool.query(
+    `
+    delete from payment_requests where payment_request_id = $1
+    `,
+    [requestId],
+  );
+
+  redirect("/paymentSetup/requests");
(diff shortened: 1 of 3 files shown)


apps/payments/app/[locale]/(hosted)/paymentSetup/requests/[requestId]/page.tsx
```diff
@@ -49,11 +77,27 @@ export const RequestDetails = async ({ requestId }: { requestId: string }) => {
...
+        >
+          <Link href={`/paymentSetup/edit/${requestId}`}>
+            <button className="govie-button govie-button--primary">
+              {tCommon("edit")}
+            </button>
+          </Link>
+          <form action={deletePR}>
+            <button
+              className="govie-button govie-button--tertiary"
+              // I'd like to add a tooltip here to explain why the button is disabled, but tooltips are not working for me
```
Comment (alfonsograziano) [inline apps/payments/app/[locale]/(hosted)/paymentSetup/requests/[requestId]/page.tsx:94]: @Thurfjell do you think we should copy the JS from the Design System, or shall I just create a new Tooltip custom component outside of the DS?

peschina (approved):
Let's see what Ludwig says on the tooltip. Apart from that, LGTM

alfonsograziano:
I created the client component for tooltip since the one from the DS doesn't work as intended in our environment

PR #33, opened 2024-03-22
Fix release connection

peschina:


apps/payments/app/[locale]/(embedded)/paymentRequest/pay/page.tsx
```diff
@@ -266,6 +266,7 @@ export default async function Page(props: Props) {
                 )}
               />
             </div>
+            <hr className="govie-section-break govie-section-break--visible"></hr>
```
Comment (peschina) [inline apps/payments/app/[locale]/(embedded)/paymentRequest/pay/page.tsx:269]: Adding this here for now since they're all hardcoded, but we should probably refactor it when adding the next provider so it's a bit more flexible

alfonsograziano (approved):

apps/payments/app/[locale]/(embedded)/paymentRequest/pay/page.tsx
Comment (Thurfjell) [reply apps/payments/app/[locale]/(embedded)/paymentRequest/pay/page.tsx:269]: This is so confusing

Thurfjell (approved):

PR #34, opened 2024-03-22
Build Stripe instance at runtime

alfonsograziano:
Right now, if I don't have the `STRIPE_SECRET_KEY` created, even if I use TrueLayer the code explodes since it requires the variable to be set when we import the module.

By creating the Stripe instance only when needed, we fix this env variable check


peschina (approved):
LGTM

PR #37, opened 2024-03-22
Allow Edit providers

alfonsograziano:


apps/payments/app/[locale]/(hosted)/paymentSetup/requests/[requestId]/page.tsx
```diff
@@ -118,14 +118,16 @@ export const RequestDetails = async ({ requestId }: { requestId: string }) => {
...
-        {details.providers.map(({ provider_name, provider_type }) => (
-          <div className="govie-summary-list__row">
-            <dt className="govie-summary-list__key">
-              {t(`form.paymentProvider.${provider_type}`)}
-            </dt>
-            <dt className="govie-summary-list__value">{provider_name}</dt>
-          </div>
-        ))}
+        {details.providers
+          .filter((p) => p.enabled)
```
Comment (alfonsograziano) [inline apps/payments/app/[locale]/(hosted)/paymentSetup/requests/[requestId]/page.tsx:122]: We can also do the filter on the query side now that I'm reviewing this

apps/payments/app/[locale]/(hosted)/paymentSetup/edit/[request_id]/page.tsx
```diff
@@ -45,6 +91,36 @@ async function editPayment(
...
+    );
+
+    await updateProvider(
+      client,
+      paymentRequestId,
+      providers.find((provider) => provider.provider_type === "stripe")
+        ?.provider_id,
+      formData.get("stripe-account")?.toString(),
+    );
+
```
Comment (peschina) [inline apps/payments/app/[locale]/(hosted)/paymentSetup/edit/[request_id]/page.tsx:123]: Let's remove the TODO comment below

peschina (commented):
I think this way we're always going to execute a query to update the provider even if there is no change in the provider. Can we make sure we do only if a provider has actually been changed?

alfonsograziano:
> I think this way we're always going to execute a query to update the provider even if there is no change in the provider. Can we make sure we do only if a provider has actually been changed?

I just did a double-check with all the cases. From what I can see, it looks like we're not executing queries unless is not necessary

peschina (approved):
LGTM

package.json
```diff
@@ -16,8 +16,9 @@
     "start": "npm run start:docker && concurrently \"npm run start:services\" \"sleep 5 && npm run migrate\"",
     "reset-docker": "docker-compose down && docker system prune -a -f && docker builder prune -f && docker image prune -f",
     "test": "echo \"Error: no test specified\" && exit 1",
-    "prepare": "(command -v husky && husky) || true",
```
williamwd, line 19: This line is important to verify if husky is available at build time. Can you please revert it?

PR #38, opened 2024-03-22
E2e Auth setup

peschina:
Added a setup project that performs basic authentication and set shared storage between tests. With this config the setup  project will always run and authenticate before all the tests. The setup project is run only once per project.
This is currently working as expected with chromium.
Parallel execution will be dealt with in a separate pr.


apps/payments/e2e/providers/providers.spec.ts
```diff
@@ -0,0 +1,21 @@
+import { test } from "@playwright/test";
+
+test("should be able to add a bank transfer provider", async ({ page }) => {
```
peschina, line 3: This test will be refined in a separate PR dedicated to testing providers. It's just here to prove that with this setup all tests will be already authenticated

alfonsograziano (approved):
LGTM - for now is fine. When we put this on the CI, we will make the necessary adjustments if needed

PR #27, opened 2024-03-20
Apply for jobseeker's allowance flow

GiuliaTeggi:
Add workflow to apply for jobseeker's allowance. 

⚠️ Right now there are no tests for the web app. I'll update the playwright config (atm it's the default one) and add some tests for the workflows we have in a follow-up pr.

Recording: 

https://github.com/ogcio/life-events/assets/30217557/5e0b2a9c-6c27-4cb0-b40d-b74cf2b446b3

PR #41, opened 2024-03-22
Update web app styling and pages

GiuliaTeggi:
- Add landing pages for each event (birth, health, death etc) with 'coming soon' actions
- Update styling 
- Move some components to `components` folders

In a following pr I'll add feature flags for each menu item so that they can be hidden or shown.

Screenshot 2024-03-22 at 17 44 05

Screenshot 2024-03-25 at 09 32 36


apps/web/app/[locale]/Header.tsx
```diff
@@ -1,149 +0,0 @@
-import Link from "next/link";
```
Comment (GiuliaTeggi) [inline apps/web/app/[locale]/Header.tsx:1]: This component was just moved and updated, not deleted

apps/web/app/utils/routes.ts
```diff
@@ -151,7 +151,75 @@ export const driving = {
     },
   },
 };
-export const employment = { slug: "employment" };
+export const employment = {
```
GiuliaTeggi, line 154: This is simply copied from #27 - which should be merged before this pr.

Thurfjell (approved):
Lots of this is gonna need to be tweaked when we start consider workflows potentially coming from the messaging system. 

Which workflows should be static and which are more dynamic? 🤷 

This is good stuff regardless. Strong work.

GiuliaTeggi:
> Lots of this is gonna need to be tweaked when we start consider workflows potentially coming from the messaging system.
> 
> Which workflows should be static and which are more dynamic? 🤷
> 
> This is good stuff regardless. Strong work.

Yes absolutely, this is really only to have some landing pages to show in the demo I guess

PR #42, opened 2024-03-25
Run E2E tests in parallel

peschina:
Add a fixture for `test`, to authenticate once per worker by setting  a custom `storageState`. Tests run by the same worker will share authentication by default. This way we'll be able to run multiple Playwright projects in parallel


apps/payments/e2e/fixtures/test.ts
```diff
@@ -0,0 +1,43 @@
...
+const baseURL = process.env.HOST_URL ?? "http://localhost:3001";
+const loginUrl = process.env.LOGIN_URL ?? "http://localhost:8000/static/login/";
+
+export * from "@playwright/test";
+export const test = baseTest.extend<{}, { workerStorageState: string }>({
+  storageState: ({ workerStorageState }, use) => use(workerStorageState),
+
+  workerStorageState: [
+    async ({ browser }, use) => {
+      const id = test.info().parallelIndex;
```
peschina, line 14: We use this to identify the worker

apps/payments/e2e/fixtures/test.ts
```diff
@@ -0,0 +1,43 @@
...
+        test.info().project.outputDir,
+        `.auth/${id}.json`,
+      );
+
+      if (fs.existsSync(fileName)) {
+        await use(fileName);
+        return;
+      }
+
+      const page = await browser.newPage({ storageState: undefined });
```
peschina, line 25: clean up storage state before authenticating our user

apps/payments/playwright.config.ts
```diff
@@ -17,49 +17,35 @@ export default defineConfig({
...
+      },
+    },
+    {
+      name: "Desktop Safari",
+      use: {
+        ...devices["Desktop Safari"],
+      },
+    },
+    {
+      name: "Mobile Chrome",
```
alfonsograziano, line 39: Do we want to run the entire test suite on Mobile? I expect just a few pages (the user-facing ones) to be available on Mobile. 

Probably, the majority of the application, on the Public Servant side, will be optimized just to be used on Desktop (something we will discuss with Tamara, I guess)

alfonsograziano (approved):
LGTM

apps/payments/playwright.config.ts
peschina: makes sense, let's reiterate on this

PR #46, opened 2024-03-25
Add feature flags to main menu

GiuliaTeggi:
### Description

- Setting up default feature flags 
- Adding FF checks on menu items
 
## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [ ] **New feature**
- [ ] **Dev change**

### Checklist:

### Screenshots:

N/A


packages/feature-flags/utils.ts
```diff
@@ -1,12 +1,52 @@
...
+  housing: {
+    enabled: true,
+  },
+  death: {
+    enabled: true,
+  },
+};
+
+export async function isFeatureFlagEnabled(
+  name: string,
```
GiuliaTeggi, line 35: I have typed this as keyof typeof defaultFeatureFlags because I can create more feature flags that gets saved to my local DB and that are not amongst the default FF...don't think we need to account for all these cases right now. We can get back to this once we have some seeded feature flags and only the DB is the source of truth.

Thurfjell: I'd rather have all flags sourcing from the object tbh. Get the type and assure that there aren't any rogue flags.

Or we could just string it!  😂

Thurfjell (approved):

PR #47, opened 2024-03-25
Link providers to Transaction

alfonsograziano:
### Ticket:

- There is no ticket for this

### Description

We need to have a reference of the provider inside each transaction in order to know how the user decided to pay.
Also, the "Assign to a user" feature has been temporarily removed since will need a different implementation to fit the new requirements (like user must not be logged in to use the system)

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**

### Screenshots:

Screenshot 2024-03-25 at 19 58 18


peschina (approved):
LGTM

PR #64, opened 2024-03-28
Fixed automatic logout when on welcome page

SamSalvatico:
### Description

Next.js behaviour is different in production than in the development environment
The prefetch feature in link [only works in production](https://nextjs.org/docs/app/api-reference/components/link#prefetch) 🤯 , then, after login, in the deployed environment, was automatically logging out because of prefetch enabled (as default) on the logout link

## Type

- [ ] **Dependency upgrade**
- [x] **Bug fix**
- [ ] **New feature**
- [ ] **Dev change**

### DIFF
--- a/apps/messaging/app/[locale]/Header.tsx
+++ b/apps/messaging/app/[locale]/Header.tsx
@@ -136,7 +136,7 @@ export default () => {
             </div>
             <UserIcon />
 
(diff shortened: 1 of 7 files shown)


alfonsograziano (approved):
LGTM

PR #71, opened 2024-03-29
PAyments: Set var in dev.yml

SamSalvatico:
### Description

Given that NextJs needs an env variable during build time, we have to set it in the pipeline variables declaration

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [ ] **New feature**
- [x] **Dev change**


alfonsograziano (approved):
LGTM

PR #75, opened 2024-04-02
feat: create and list providers api

nnorbert:
### Ticket:

- [15123 ](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/15123)

### Description

Create backend API for creating and listing the providers.

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [X] **New feature**
- [ ] **Dev change**

### Checklist:

### Screenshots:

N/A

### DIFF
--- a/apps/payments/app/[locale]/(hosted)/paymentSetup/providers/add-banktransfer/page.tsx
+++ b/apps/payments/app/[locale]/(hosted)/paymentSetup/providers/add-banktransfer/page.tsx
@@ -15,10 +15,10 @@ export default async () => {
     await buildApiClient(userId).providers.apiV1ProvidersPost({
       name: formData.get("provider_name") as string,
       type: "banktransfer",
-      providerData: {
-        sortCode: formData.get("sort_code") as string,
-        accountNumber: formData.get("account_number") as string,
-        accountHolderName: formData.get("account_holder_name") as string,
+      data: {
+        sortCode: formData.get("sort_code"),
+        accountNumber: formData.get("account_number"),
+        accountHolderName: formData.get("account_holder_name"),
       },
     });
 
--- a/apps/payments/app/[locale]/(hosted)/paymentSetup/providers/add-banktransfer/page.tsx
+++ b/apps/payments/app/[locale]/(hosted)/paymentSetup/providers/add-banktransfer/page.tsx
@@ -1,7 +1,7 @@
 import { getTranslations } from "next-intl/server";
 import { PgSessions } from "auth/sessions";
 import { redirect } from "next/navigation";
-import { pgpool } from "../../../../../dbConnection";
+import buildApiClient from "../../../../../../client/index";
 import BankTransferFields from "./BankTransferFields";
 
 export default async () => {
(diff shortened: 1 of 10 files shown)


apps/payments-backend/routes/providers/index.ts
```diff
@@ -1,28 +1,47 @@
...
     providerData: Type.Object({}),
+    status: Type.Union([
+      Type.Literal("connected"),
+      Type.Literal("disconnected"),
+    ]),
   }),
 ]);
 
-export type ProviderType = Static<typeof Provider>;
+export const ProvidersList = Type.Union([Type.Array(Provider)]);
```
alfonsograziano, line 33: Do we need to export all this stuff? Are we importing it somewhere?

apps/payments-backend/routes/providers/index.ts
```diff
@@ -48,4 +67,58 @@ export default async function providers(app: FastifyInstance) {
...
+        },
+      },
+    },
+    async (request, reply) => {
+      const userId = request.user?.id;
+
+      const result = await app.pg.query(
+        `
+          SELECT
+            provider_id as "providerId",
```
alfonsograziano, line 110: Since the table is "payment_providers", shall we simplify the names using "id", "name", "type", "data"? We may want to do that in a different PR though, not related to this one

apps/payments/app/[locale]/(hosted)/paymentSetup/providers/add-banktransfer/page.tsx
```diff
@@ -11,28 +11,16 @@ export default async () => {
...
-          accountNumber,
-          accountHolderName,
-        },
-      ],
-    );
+    await buildApiClient(userId).providers.apiV1ProvidersPost({
+      name: formData.get("provider_name") as string,
+      type: "banktransfer",
+      providerData: {
+        sortCode: formData.get("sort_code") as string,
```
Comment (alfonsograziano) [inline apps/payments/app/[locale]/(hosted)/paymentSetup/providers/add-banktransfer/page.tsx:19]: Since this is a generic object without further type enforcing, for now I guess we don't need to add `as string`

alfonsograziano (approved):
LGTM - apart from some optional comments

apps/payments-backend/routes/providers/index.ts
nnorbert: Not at this moment. I will update it.

apps/payments-backend/routes/providers/index.ts
nnorbert: Definitely, yes. Because the entity is provider, the properties inside should not repeat the entity's name. So I also think, we should use "id", "name", and so on.

apps/payments/app/[locale]/(hosted)/paymentSetup/providers/add-banktransfer/page.tsx
Comment (nnorbert) [reply apps/payments/app/[locale]/(hosted)/paymentSetup/providers/add-banktransfer/page.tsx:19]: Sure, I will update it.

apps/payments-backend/routes/providers/index.ts
nnorbert: Updated

apps/payments-backend/routes/providers/index.ts
nnorbert: Updated

apps/payments/app/[locale]/(hosted)/paymentSetup/providers/add-banktransfer/page.tsx
Comment (nnorbert) [reply apps/payments/app/[locale]/(hosted)/paymentSetup/providers/add-banktransfer/page.tsx]: Updated

alfonsograziano (approved):

PR #76, opened 2024-04-02
Allow payment method choice within an iframe

alfonsograziano:
### Description

By changing the payment method selection from a series of links to a form with multiple radio buttons, I accidentally broke the iframe flow due to a missing parent redirect. This fixes the issue

## Type

- [ ] **Dependency upgrade**
- [X] **Bug fix**
- [ ] **New feature**
- [ ] **Dev change**


azure_pipelines.yml
```diff
@@ -17,10 +17,16 @@ parameters:
   - name: destroy
     type: boolean
     default: false
-  
+
   - name: envBranches
     type: object
-    default: ["refs/heads/dev", "refs/heads/sta", "refs/heads/uat", "refs/heads/feat/improve-pipeline"]
+    default:
```
alfonsograziano, line 23: Sorry for all the changes here, must be the prettier plugin with the format on save

azure_pipelines.yml
```diff
@@ -46,106 +52,106 @@ stages:
...
-            action: 'Run a Docker command'
-            customCommand: 'load -i $(Pipeline.Workspace)/docker-design-system-$(Build.BuildId)/image_design_system_$(Build.BuildId).tar'
-        - task: Docker@2
-          displayName: Build - ${{ serviceName }}
-          inputs:
-            command: build
-            repository: life-events-${{ serviceName }}
-            dockerfile: apps/${{ serviceName }}/Dockerfile
-            buildContext: '$(Build.SourcesDirectory)'
-            arguments: '--build-arg NEXT_PUBLIC_STRIPE_PUBLISHABLE_KEY=$(nextPublicStripePublishableKey)'
```
SamSalvatico, line 118: You have to set `NEXT_PUBLIC_HOST_URL` also here if you need to use it in the Dockerfile

pipeline-variables/dev.yml
```diff
@@ -2,4 +2,5 @@ variables:
   awsServiceConnection: life-events-dev-ecr
   awsRegion: eu-west-1
   ecrEndpoint: 730335224023.dkr.ecr.eu-west-1.amazonaws.com
-  nextPublicStripePublishableKey: pk_test_51OzFkUK1VL0xyyr2E3EvcEmzsAGLowKH3jlM4g3AhQwUjhlpxQvfjZizc5GRSAnNzf8HHwTdzhz2rRYS69nooxtg00ZF7kRvyN
\ No newline at end of file
+  nextPublicStripePublishableKey: pk_test_51OzFkUK1VL0xyyr2E3EvcEmzsAGLowKH3jlM4g3AhQwUjhlpxQvfjZizc5GRSAnNzf8HHwTdzhz2rRYS69nooxtg00ZF7kRvyN
+  nextPublicHostUrl: http://payments.dev.blocks.gov.ie
```
SamSalvatico, line 6: In the .env file you've set a trailing slash at the end of the URL, just ensure that it would work also without it (or just add it here too 😁 )

SamSalvatico (changes requested):

azure_pipelines.yml
alfonsograziano: Fixed, thanks!

pipeline-variables/dev.yml
alfonsograziano: Good catch, to keep it consistent I added the slask

SamSalvatico (approved):
LGTM

azure_pipelines.yml
williamwd: No problem. I'll revert some things just so it stays consistent with the Azure documentation.

PR #84, opened 2024-04-03
Fix types

alfonsograziano:
### Description

Fixing a type for params on the APIs

## Type

- [ ] **Dependency upgrade**
- [x] **Bug fix**
- [ ] **New feature**
- [ ] **Dev change**

### DIFF
--- a/apps/payments-api/routes/providers/index.ts
+++ b/apps/payments-api/routes/providers/index.ts
@@ -28,6 +28,11 @@ type ProvidersListType = Static<typeof ProvidersList>;
 const UpdateProvider = Type.Omit(Provider, ["id", "type"]);
 type UpdateProviderType = Static<typeof UpdateProvider>;
 
+const ParamsWithProviderId = Type.Object({
+  providerId: Type.String(),
+});
+type ParamsWithProviderId = Static<typeof ParamsWithProviderId>;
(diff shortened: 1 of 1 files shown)


apps/payments-api/routes/providers/index.ts
```diff
@@ -28,6 +28,11 @@ type ProvidersListType = Static<typeof ProvidersList>;
 const UpdateProvider = Type.Omit(Provider, ["id", "type"]);
 type UpdateProviderType = Static<typeof UpdateProvider>;
 
+const ParamsWithProviderId = Type.Object({
+  providerId: Type.String(),
+});
+type ParamsWithProviderId = Static<typeof ParamsWithProviderId>;
```
nnorbert, line 34: The rest of the types have 'Type' suffix. Just asking if you intentionally skipped or just a mistake?

nnorbert (approved):
Just one question. LGTM!

apps/payments-api/routes/providers/index.ts
nnorbert: Indeed, that is very right. I just was curious if you wish to maintain the format, which previously was used. But sure, we can update the rest of the types. Thanks!

PR #85, opened 2024-04-03
Payments - introduce eslint

alfonsograziano:


apps/payments/app/[locale]/(hosted)/paymentSetup/PaymentSetupForm.tsx
```diff
@@ -76,7 +76,7 @@ export default async function ({
             (p) => p.provider_type === providerType,
           );
           return (
-            <div className="govie-form-group">
+            <div className="govie-form-group" key={index}>
```
Comment (nnorbert) [inline apps/payments/app/[locale]/(hosted)/paymentSetup/PaymentSetupForm.tsx:79]: As far as I know, indexes are not recommended for use as keys because, based on this key, React keeps tracking the items. The index will be the same if two items are changed in order.

However, I think this is a small list, right? Then maybe it is not that important in our case.

nnorbert (approved):
Only one minor comment. It is discussable if it even worth the effort. In rest it looks good!

apps/payments/app/[locale]/(hosted)/paymentSetup/PaymentSetupForm.tsx
Comment (alfonsograziano) [reply apps/payments/app/[locale]/(hosted)/paymentSetup/PaymentSetupForm.tsx:79]: Using an index is perfectly fine in case you know that the list is not dynamic. In this case, we're only generating three items, one for each provider, and this will never change during the lifetime of the application. Should be good enough :)

PR #89, opened 2024-04-05
feat: create payment request API

nnorbert:
### Ticket:

- [15123](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/15123)

### Description

Please include a summary of the changes and the related issue.

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [X] **New feature**
- [ ] **Dev change**

### Checklist:

### Screenshots:

N/A

### DIFF
--- a/apps/payments-api/routes/paymentRequests/index.ts
+++ b/apps/payments-api/routes/paymentRequests/index.ts
@@ -50,12 +50,7 @@ const CreatePaymentRequest = Type.Object({
   redirectUrl: Type.String(),
   allowAmountOverride: Type.Boolean(),
   allowCustomAmount: Type.Boolean(),
-  accounts: Type.Array(
-    Type.Object({
-      account: Type.String(),
-      enabled: Type.Boolean(),
-    }),
-  ),
+  providers: Type.Array(Type.String()),
 });
 type CreatePaymentRequest = Static<typeof CreatePaymentRequest>;
 
@@ -221,7 +216,7 @@ export default async function paymentRequests(app: FastifyInstance) {
         redirectUrl,
         allowAmountOverride,
         allowCustomAmount,
-        accounts,
+        providers,
       } = request.body;
 
       try {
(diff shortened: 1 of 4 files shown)


apps/payments-api/routes/paymentRequests/index.ts
```diff
@@ -42,6 +42,23 @@ const ParamsWithPaymentRequestId = Type.Object({
...
 
+const CreatePaymentRequest = Type.Object({
+  title: Type.String(),
+  description: Type.String(),
+  reference: Type.String(),
+  amount: Type.Number(),
+  redirectUrl: Type.String(),
+  allowAmountOverride: Type.Boolean(),
+  allowCustomAmount: Type.Boolean(),
+  accounts: Type.Array(
```
alfonsograziano, line 53: this is not an account, but rather a `provider`

apps/payments-api/routes/paymentRequests/index.ts
```diff
@@ -42,6 +42,23 @@ const ParamsWithPaymentRequestId = Type.Object({
...
+    Type.Object({
+      account: Type.String(),
+      enabled: Type.Boolean(),
```
alfonsograziano, line 56: Once we create the provider <-> payment request relation, by default it must be enabled, we can avoid sending this param here and we can set if enabled = true directly in the query

apps/payments/app/[locale]/(hosted)/paymentSetup/create/page.tsx
```diff
@@ -1,158 +1,56 @@
...
-  const stripeAccount = formData.get("stripe-account")?.toString();
-  const worldpayAccount = formData.get("worldpay-account")?.toString();
-
-  if (
-    !openBankingAccount &&
-    !bankTransferAccount &&
-    !stripeAccount &&
-    !worldpayAccount
-  ) {
+  const accounts: Account[] = [
```
Comment (alfonsograziano) [inline apps/payments/app/[locale]/(hosted)/paymentSetup/create/page.tsx:15]: Same here, we need to rename this to `provider`. We used the word "account" early once we didn't have yet the glossary.

apps/payments-api/routes/paymentRequests/index.ts
nnorbert: For one of the providers it was hardcoded to false. So that's why I am sending it. This will make the code simpler.

alfonsograziano: I think WorldPay has been hardcoded to false since the integration is not ready yet. For now, you can just pretend WorldPay doesn't exist as a provider and don't send it from the frontend in the array

apps/payments/app/[locale]/(hosted)/paymentSetup/create/page.tsx
```diff
@@ -1,158 +1,44 @@
...
-    [userId, providerType],
-  );
-
-  if (!accountsQueryResult.rowCount) {
-    return [];
-  }
-
-  return accountsQueryResult.rows;
-}
+type Account = {
```
Comment (alfonsograziano) [inline apps/payments/app/[locale]/(hosted)/paymentSetup/create/page.tsx:7]: Do we need this? :)

alfonsograziano (approved):
Apart from an unused type, LGTM. Great :)

PR #90, opened 2024-04-05
Setup profile app

GiuliaTeggi:
### Description

Setting up the profile app. This is a new app running locally at port `3003`. 

As part of the setup, a new `profile` database is created, containing two tables `user_details` and `addresses` (⚠️ these are only needed for demo purposes - eventually all user data should come from external apis).

Large parts of the new files are copied from the `web` app setup.

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**

### Checklist:

### Screenshots:

Screenshot 2024-04-08 at 14 38 01


apps/profile/app/components/FeedbackBanner.tsx
```diff
@@ -0,0 +1,26 @@
+import { useTranslations } from "next-intl";
```
GiuliaTeggi, line 1: The following components are copied from the `web` app

apps/profile/migrations/sqls/20240405132138-init-up.sql
```diff
@@ -0,0 +1,30 @@
...
+    gender TEXT NOT NULL,
+    phone TEXT NOT NULL, 
+    email TEXT NOT NULL,
+    created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    updated_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    primary key(user_id)
+);
+
+CREATE TABLE IF NOT EXISTS addresses (
+    address_id SERIAL PRIMARY KEY,
```
Thurfjell, line 17: why serial

apps/profile/migrations/sqls/20240405132138-init-up.sql
```diff
@@ -0,0 +1,30 @@
...
+    user_id UUID NOT NULL,
+    address_line1 TEXT NOT NULL,
+    address_line2 TEXT,
```
Thurfjell, line 20: Is this verified to always be at least one, but at most two? What is an address line?

apps/profile/migrations/sqls/20240405132138-init-up.sql
```diff
@@ -0,0 +1,30 @@
...
+    city TEXT NOT NULL,
+    eirecode TEXT NOT NULL,
+    is_primary BOOLEAN NOT NULL DEFAULT FALSE,
+    ownership_status TEXT NOT NULL,
+    start_date DATE NOT NULL,
```
Thurfjell, line 25: Could also use timestamptz and cast to date if it's a nicer format to read.

apps/profile/migrations/sqls/20240405132138-init-up.sql
```diff
@@ -0,0 +1,30 @@
+CREATE TABLE IF NOT EXISTS user_details(
+    user_id UUID NOT NULL,
+    title TEXT NOT NULL,
+    name TEXT NOT NULL,
```
Thurfjell, line 4: What kind of name is this?

apps/profile/migrations/sqls/20240405132138-init-up.sql
```diff
@@ -0,0 +1,30 @@
...
+    surname TEXT NOT NULL,
+    date_of_birth DATE NOT NULL,
+    ppsn TEXT NOT NULL,
+    gender TEXT NOT NULL,
+    phone TEXT NOT NULL, 
+    email TEXT NOT NULL,
```
Thurfjell, line 10: Can a user have multiple mails/phones associated with a profile?

apps/profile/migrations/sqls/20240405132138-init-up.sql
```diff
@@ -0,0 +1,30 @@
...
+    address_line2 TEXT,
+    city TEXT NOT NULL,
+    eirecode TEXT NOT NULL,
+    is_primary BOOLEAN NOT NULL DEFAULT FALSE,
+    ownership_status TEXT NOT NULL,
+    start_date DATE NOT NULL,
+    end_date DATE,
+    created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    updated_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    FOREIGN KEY (user_id) REFERENCES user_details(user_id)
```
Thurfjell, line 29: Should it be named user_addresses due to this constraint?

init.sql
```diff
@@ -3,6 +3,7 @@ CREATE DATABASE life_events;
 CREATE DATABASE messaging;
 CREATE DATABASE payments;
 CREATE DATABASE shared;
+CREATE DATABASE profile;
```
Thurfjell, line 6: Not profiles?

apps/profile/migrations/sqls/20240405132138-init-up.sql
```diff
@@ -0,0 +1,30 @@
+CREATE TABLE IF NOT EXISTS user_details(
+    user_id UUID NOT NULL,
```
Thurfjell, line 2: We are always going to provide this right? Why not add a default?

apps/profile/migrations/sqls/20240405132138-init-up.sql
GiuliaTeggi: It's literally this: 
Screenshot 2024-04-08 at 14 53 59

I can review this once I start using this table, but I thought that usually one line is mandatory whereas the second is not.

Thurfjell: Gotcha, probably not worth a link table for that stuff. Could use a text array as well? You decide should be fine to change to whatever until we go live.

init.sql
GiuliaTeggi: Because the name of the app is `profile`, so this refers to that

apps/profile/migrations/sqls/20240405132138-init-up.sql
GiuliaTeggi: I'll change it to `uuid`

apps/profile/migrations/sqls/20240405132138-init-up.sql
Thurfjell: Might be worth checking out? Not a big deal.

apps/profile/migrations/sqls/20240405132138-init-up.sql
GiuliaTeggi: Having a separate table for addresses makes it easier to add multiple addresses for the same user, as that's what I need to do for the first iteration of the 'about me' page. Yes agreed the two lines could just be part of the same text string eventually.

Thurfjell (approved):

PR #93, opened 2024-04-05
Add public api endpoint env var

GiuliaTeggi:
### Ticket:

- [15009](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/15009)

### Description

Adding env var needed to fetch the timeline

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [ ] **New feature**
- [ ] **Dev change**

### Checklist:

### Screenshots:

N/A


williamwd:
The context is that these environment variables are needed at the frontend. This is not an issue for local development, but running in production mode requires these variables to be available at build time.
More info: https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables#bundling-environment-variables-for-the-browser

PR #95, opened 2024-04-05
Logto integration

alfonsograziano:
### Ticket:

### Description

Please include a summary of the changes and the related issue.

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [ ] **New feature**
- [ ] **Dev change**

### Checklist:

### Screenshots:

N/A


apps/payments-api/plugins/logtoAuth.ts
```diff
@@ -0,0 +1,63 @@
...
+  requiredPermissions: Permission[],
+) => {
+  try {
+    const token = extractBearerTokenFromHeaders(req);
+    console.log(token);
+    const { payload } = await jwtVerify(token, jwks, {
+      // Expected issuer of the token, issued by the Logto server
+      issuer: "http://localhost:3301/oidc",
+      // Expected audience token, the resource indicator of the current API
+      audience: "http://localhost:8001",
```
relvao, line 30: hum... would this mean that each token would only be valid for a specific API? I'm not sure I follow this

alfonsograziano: On the frontend, we're issuing a token for a specific resource (payments api), so I guess that yes, the token generated is valid for a single resource. Of course we can generate multiple tokens for multiple resources

SamSalvatico: As Alfonso wrote the LogTo token management for API resource works giving one token for each single resource we need, that's because the permissions are related to an API resource.
Payments will have its own scopes, messaging its own and so on. But, but: a role is not related to a resource, then it can have permissions for different resources at the same time. In this way, with one role, a "normal" user can have access to multiple resources
https://docs.logto.io/docs/recipes/rbac/protect-resource/

PR #102, opened 2024-04-09
Bootstrap building blocks sdk

alfonsograziano:
### Description

We want to create a single SKD for the building blocks

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [ ] **New feature**
- [X] **Dev change**


apps/payments-api/package.json
```diff
@@ -8,7 +8,8 @@
     "start": "node --import tsx index.ts",
     "dev": "nodemon",
     "generate-client": "openapi-generator-cli generate -i openapi-definition.yml -g typescript-axios -o ../payments/client/autogenerated",
-    "lint": "eslint . --ext .ts"
+    "lint": "eslint . --ext .ts",
+    "generate-ts-client": "npx openapi-typescript ./openapi-definition.yml -o ../../packages/building-blocks-sdk/services/payments/schema.d.ts"
```
alfonsograziano, line 12: The generate command has been created here, since the SDK must not be aware of how the content of the SDK is created (autogenerated or hand-written)

apps/payments/app/[locale]/(hosted)/paymentSetup/providers/ProvidersList.tsx
```diff
@@ -2,15 +2,14 @@ import Link from "next/link";
...
 
 export default async () => {
   const t = useTranslations("PaymentSetup.Providers.table");
   const { userId } = await PgSessions.get();
-  const providers = (await buildApiClient(userId).providers.apiV1ProvidersGet())
-    .data;
+  const { data: providers } = await new Payments(userId).getProviders();
 
-  if (providers.length === 0) {
+  if (providers && providers.length === 0) {
```
Comment (temsa) [inline apps/payments/app/[locale]/(hosted)/paymentSetup/providers/ProvidersList.tsx:12]:
```suggestion
  if (providers?.length === 0) {
```

packages/building-blocks-sdk/services/payments/schema.d.ts
temsa, line 1: If this file is generated, should it be actually committed?

apps/payments/app/[locale]/(hosted)/paymentSetup/providers/ProvidersList.tsx
```diff
@@ -36,22 +35,23 @@ export default async () => {
...
-            </td>
-            <td className="govie-table__cell govie-table__cell--vertical-centralized govie-body-s">
-              {provider.name}
-            </td>
-            <td className="govie-table__cell govie-table__cell--vertical-centralized govie-body-s govie-table__header--numeric">
-              <Link href={`providers/${provider.id}`}>{t("edit")}</Link>
-            </td>
-          </tr>
-        ))}
+        {providers &&
```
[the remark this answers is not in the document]
Comment (alfonsograziano) [reply apps/payments/app/[locale]/(hosted)/paymentSetup/providers/ProvidersList.tsx:38]: Definitely! Right now we're not managing Empty/Error states yet, but it totally makes sense to have it :D

packages/building-blocks-sdk/services/payments/schema.d.ts
alfonsograziano: The idea is that the sdk is not aware that the schema file has been autogenerated. At this moment we decided to autogenerate it, but in the future, we may want to manually override it, or even use custom types. In fact, the `building-blocks-sdk` package doesn't contain the logic to generate the types. 

At least, this is the initial idea, but I'm open to discussion :D

Thurfjell (approved):

PR #110, opened 2024-04-10
Make payment working even if unauthenticated

alfonsograziano:
### Description

Allow users to make a payment in incognito mode. This creates some vulnerabilities we will have to address

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [X] **New feature**
- [ ] **Dev change**


apps/payments-api/routes/paymentRequests/index.ts
```diff
@@ -149,6 +149,75 @@ export default async function paymentRequests(app: FastifyInstance) {
     },
   );
 
+  app.get<{
+    Reply: PaymentRequestDetails | Error;
+    Params: ParamsWithPaymentRequestId;
+  }>(
+    "/:requestId/summary",
```
alfonsograziano, line 156: This is a `public` summary of the request containing only the publicly accessible info. Doesn't require any auth

alfonsograziano:
Closing due a change in the product direction

PR #112, opened 2024-04-11
feat: status normalisation

nnorbert:
Transaction status normalisation

### DIFF
--- a/apps/payments-api/routes/transactions/index.ts
+++ b/apps/payments-api/routes/transactions/index.ts
@@ -4,6 +4,8 @@ import {
   CreateTransactionBody,
   ParamsWithTransactionId,
   TransactionDetails,
+  TransactionStatuses,
+  TransactionStatusesEnum,
   UpdateTransactionBody,
 } from "../../types/schemaDefinitions";
 import { Type } from "@sinclair/typebox";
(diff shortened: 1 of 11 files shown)


apps/payments-api/routes/transactions/index.ts
```diff
@@ -125,14 +127,15 @@ export default async function transactions(app: FastifyInstance) {
...
-          values ($1, $2, $3, $4, 'pending', now(), now(), $5, $6)
+          values ($1, $2, $3, $4, $5, now(), now(), $6, $7)
           returning transaction_id as "transactionId";
         `,
         [
           paymentRequestId,
           extPaymentId,
           integrationReference,
           amount,
+          TransactionStatusesEnum.Pending,
```
alfonsograziano, line 138: Once we do the insert, shouldn't the status be "Initialized"?

alfonsograziano (changes requested):
As fare as I can see, here has not been updated the manual bank transfer flow with the correct status, right?

apps/payments-api/routes/transactions/index.ts
nnorbert: However, this part was changed a bit, status is an optional parameter now. If you omit to send status, it will create as Pending by default. OpenBanking and Stripe will have Pending status by default, only Manual transfer use the initiated status. However, it can be changed to insert Initiated by default, and send Pending in case of OB and Stripe.

alfonsograziano (approved):

PR #113, opened 2024-04-11
Implenent IBAN field and change currency to EUR

nnorbert:
Implenent IBAN field and change currency to EUR


alfonsograziano (approved):
LGTM

PR #121, opened 2024-04-16
Allow only publicServant to access to payments backoffice

alfonsograziano:
### Ticket:

- [15516](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/15516)

### Description

In preparation for a different UI between public servants and citizens, we block the access to the backoffice pages in case of missing permissions

## Type

- [ ] **Dependency upgrade**
- [x] **Bug fix**
- [ ] **New feature**
- [ ] **Dev change**

### DIFF
--- a/apps/payments/app/[locale]/(hosted)/admin/page.tsx
+++ b/apps/payments/app/[locale]/(hosted)/admin/page.tsx
@@ -1,0 +1,12 @@
+import { notFound, redirect, RedirectType } from "next/navigation";
+import { routeDefinitions } from "../../../routeDefinitions";
+import { PgSessions } from "auth/sessions";
+
+export default async () => {
+  const { publicServant } = await PgSessions.get();
+
+  if (publicServant)
+    return redirect(routeDefinitions.paymentSetup.slug, RedirectType.replace);
+
+  return notFound();
+};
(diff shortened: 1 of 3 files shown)


apps/payments/app/[locale]/(hosted)/admin/page.tsx
```diff
@@ -0,0 +1,12 @@
+import { notFound, redirect, RedirectType } from "next/navigation";
+import { routeDefinitions } from "../../../routeDefinitions";
+import { PgSessions } from "auth/sessions";
+
+export default async () => {
+  const { publicServant } = await PgSessions.get();
+
+  if (publicServant)
```
Comment (nnorbert) [inline apps/payments/app/[locale]/(hosted)/admin/page.tsx:8]: It is better to check the negative outcome first and after all the checks let the page to do its default purpose.

```suggestion
  if (!publicServant)
    return notFound();
    
return redirect(routeDefinitions.paymentSetup.slug, RedirectType.replace);  
```

nnorbert (approved):
Just one suggestion. Looks good!

apps/payments/app/[locale]/(hosted)/admin/page.tsx
Comment (alfonsograziano) [reply apps/payments/app/[locale]/(hosted)/admin/page.tsx]: Integrated, thanks :)

PR #125, opened 2024-04-17
Fix: improve dx scripts

SamSalvatico:
### Description

Added a way to run logto easily locally
Improved package scripts  to make us able to have a locally working environment in one step

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [ ] **New feature**
- [x] **Dev change**


SamSalvatico:
Please, when reviewing this, checkout this branch locally and run `npm run start` locally to check how many bug it has 😁 
Because "it's working on my machine", but.......

alfonsograziano (approved):
As a wise man once said, "It works on my machine!"

alfonsograziano:
Before merging, would be ideal to have the approval of someone who doesn't have a prior logto instance. Let's wait for another review :)

GiuliaTeggi (approved):
Tested this locally, didn't get any errors running the start command 👍  (I already had installed logto previously though)

relvao (approved):
LGTM

temsa:
I've not tried it, and it feels legit as a first step for running this locally, though it also feels a bit hacky. 

Something that feels to me "cleaner",  would it not be better (doesn't have to be right away, or in this PR):
1. to publish a docker image of logto in logto's CI with a `latest` tag ( as prod image should be immutable, it could be a "dev" version of it )
2. use the build logto image in docker-compose and force the pull to `always` in it
3. possibly use a docker-compose `profile` to run logto locally or not

SamSalvatico:
> I've not tried it, and it feels legit as a first step for running this locally, though it also feels a bit hacky.
> 
> Something that feels to me "cleaner", would it not be better (doesn't have to be right away, or in this PR):
> 
> 1. to publish a docker image of logto in logto's CI with a `latest` tag ( as prod image should be immutable, it could be a "dev" version of it )
> 2. use the build logto image in docker-compose and force the pull to `always` in it
> 3. possibly use a docker-compose `profile` to run logto locally or not

Totally agree, Tiago 
For that reason I've set a `LOGTO_IMAGE_MODE` variable to be able to use a "remote" version once is available

PR #127, opened 2024-04-17
Profile: fix locale on links and redirects

GiuliaTeggi:
### Description

This is to (hopefully - as could only repro on dev but not locally) fix the redirection to the correct `locale` pages by using the `next-inlt` `Link` and `redirect` methods, which wrap the next.js navigation ones and append the correct locale prefix to each url. See https://next-intl-docs.vercel.app/docs/routing/navigation#shared-pathnames

## Type

- [ ] **Dependency upgrade**
- [x] **Bug fix**
- [ ] **New feature**
- [ ] **Dev change**

### Checklist:

### Screenshots:

N/A


Thurfjell (approved):
This works? Magician @GiuliaTeggi 🧙

GiuliaTeggi:
> This works? Magician @GiuliaTeggi 🧙

I hope so 😄

PR #131, opened 2024-04-18
Map transaction status color

nnorbert:
Map transaction status color:

Initiated = blue
Pending = yellow
Succeeded = green
Failed and Cancelled = red


alfonsograziano (approved):
LGTM

PR #142, opened 2024-04-19
Secure providers private keys

peschina:
### Ticket:

- [15503](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/15503)

### Description

When a Stripe provider is created its secret key is now encrypted and stored in the database. When the provider details are fetched the secret key is decrypted. A symmetric key is used for encryption/decryption and it's currently stored as an environment variable. You can create a valid key locally running: `crypto.randomBytes(32).toString("base64")`.
We're still expecting stripe public key and private key in the environment variables when running the application locally when seeding the database (atm this is available for local development only). The user we currently seed is now a public servant.
In the form for editing a stripe provider the clear text private key is hidden with a password input. We can discuss if we want to add a show/hide clear text toggle in the future - this feature most probably won't be available with js disabled.

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**

### Checklist:

- [x] I have added **partial** tests that prove my fix is effective or that my feature works

### Screenshots:

N/A


alfonsograziano (approved):
LGTM - and thanks a lot for fixing the seeding script!

PR #143, opened 2024-04-19
Feat/new side menu

edge33:
### Description

Moves the life events menu to a reactive sidebar for both desktop and mobile users

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**

### Screenshots:

Screenshot 2024-04-19 at 13 33 07
Screenshot 2024-04-19 at 13 33 33


apps/web/app/[locale]/[event]/components/Menu/LifeEventsMenu.tsx
```diff
@@ -0,0 +1,82 @@
...
+import ds from "design-system";
+import styles from "./LifeEventsMenu.module.scss";
+
+const Icon = ds.Icon;
+
+type Props = {
+  options: {
+    key: string;
+    url: string;
+    icon: string;
```
Comment (alfonsograziano) [inline apps/web/app/[locale]/[event]/components/Menu/LifeEventsMenu.tsx:11]:
```
type IconProps = React.ComponentProps<typeof ds.Icon>;
icon: IconProps["icon"];
```

This approach gives type safety and allows to remove the `as any` on the bottom

apps/web/app/[locale]/[event]/page.tsx
```diff
@@ -35,11 +39,29 @@ export default async (props: web.NextPageProps) => {
 
   const Component = componentsMap[props.params.event];
 
+  // const showEventsMenu = await isFeatureFlagEnabled("eventsMenu");
```
Comment (alfonsograziano) [inline apps/web/app/[locale]/[event]/page.tsx:42]: Let's remove this comment in case you don't need it

alfonsograziano (approved):
LGTM apart from a couple of minor comments

apps/web/app/[locale]/[event]/components/TimelineWrapper.tsx
```diff
@@ -0,0 +1,43 @@
+import { AbstractIntlMessages, NextIntlClientProvider } from "next-intl";
+import ds from "design-system";
+
+import Timeline from "./Timeline";
+
+type TTimelineWrapperProps = {
```
Comment (GiuliaTeggi) [inline apps/web/app/[locale]/[event]/components/TimelineWrapper.tsx:6]: nit: is the double T needed? :)

apps/web/app/components/HamburgerIcon.tsx
```diff
@@ -0,0 +1,20 @@
+"use client";
+import ds from "design-system/";
+
+export default () => {
+  return (
+    <div
+      style={{ cursor: "pointer" }}
+      onClick={() => {
+        window.dispatchEvent(new Event("toggleSidebar"));
+      }}
```
GiuliaTeggi, line 10: I think this should be a `button`  with an aria-label instead of a `div` for better accessibility (after that I think this component could be renamed to hamburgerButton or something like that, given the icon itself is taken from the design system)

GiuliaTeggi (commented):
Looks good! I wonder if we should set the `eventsMenu` feature flag to enabled as a default now that it doesn't sit in the same place where the sidebar is?

apps/web/messages/ga.json
```diff
@@ -1,5 +1,16 @@
...
+  "EventsMenu": {
+    "events": "Events",
+    "aboutMe": "About me",
+    "birth": "Birth",
+    "health": "Health",
+    "driving": "Driving",
+    "employment": "Employment",
+    "business": "Starting a business",
+    "housing": "Housing",
+    "death": "Death"
```
GiuliaTeggi, line 14: You don't need to add this, we don't have a gaelic translation yet as you can see :D

GiuliaTeggi (approved):

PR #147, opened 2024-04-22
Allow authenticated transactions only

peschina:
### Ticket:

- [15514](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/15514)

### Description
Users are now forced to be logged in when entering the payment page.
For now name and email of the user creating the transaction are still saved in the database based on what is filled in the form at payment method selection. The form is now pre-filled with user name and email of the current authenticated user.

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**

### Checklist:

### Screenshots:

https://github.com/ogcio/life-events/assets/47299026/50291785-2d2d-4afb-a3f7-9a2431c73ba0


apps/payments/app/[locale]/(embedded)/paymentRequest/pay/SelectPaymentMethod.tsx
```diff
@@ -47,6 +50,7 @@ function getPaymentUrl({
   );
   url.searchParams.set("paymentId", paymentId);
   url.searchParams.set("integrationRef", integrationRef);
+  url.searchParams.set("userId", userId);
```
Comment (alfonsograziano) [inline apps/payments/app/[locale]/(embedded)/paymentRequest/pay/SelectPaymentMethod.tsx:53]: Do we need to pass this as url param? Can we get it directly insidie the target page with the auth package?

apps/payments/app/[locale]/(embedded)/paymentRequest/pay/SelectPaymentMethod.tsx
```diff
@@ -88,6 +94,7 @@ export default function ({
             type="text"
             id="name"
             name="name"
+            value={user.userName}
```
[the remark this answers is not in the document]
Comment (peschina) [reply apps/payments/app/[locale]/(embedded)/paymentRequest/pay/SelectPaymentMethod.tsx]: Good catch! 6144e714397bd2986650aa9120548b265636ce96

apps/payments/app/[locale]/(embedded)/paymentRequest/pay/SelectPaymentMethod.tsx
Comment (alfonsograziano) [reply apps/payments/app/[locale]/(embedded)/paymentRequest/pay/SelectPaymentMethod.tsx]: Removing the form entirely makes sense for me if we're still saving that data in the db. Not sure if we can think of a use case where the info in that form could be different from the user info, but I guess that if we will need it, we can quickly re-add the form in another PR :)

alfonsograziano (approved):
LGTM

apps/payments/app/[locale]/(embedded)/paymentRequest/pay/SelectPaymentMethod.tsx
Comment (peschina) [reply apps/payments/app/[locale]/(embedded)/paymentRequest/pay/SelectPaymentMethod.tsx]: 9166f1dcedb62788e043758b11cf6d8c9669f665

alfonsograziano (approved):

PR #149, opened 2024-04-22
Initial Logto Next.js integration

alfonsograziano:
### Description

Integrating logto under an environment variable to test it out.
Multiple PRs (also for the backend side) will follow.

This PR doesn't disable the current mock auth. This means that the user will need to login twice. Not ideal from from an UX point of view, but needed to not break the current features

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [X] **New feature**
- [ ] **Dev change**


apps/payments/libraries/auth.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  return uuid;
+}
+
+export const getUser = async () => {
+  const context = await getLogtoContext(logtoConfig, {
+    fetchUserInfo: true,
+    getAccessToken: true,
+    // The resource must have a trailing slash
+    resource: process.env.PAYMENTS_BACKEND_URL + "/",
```
SamSalvatico, line 24: Please check if `PAYMENTS_BACKEND_URL` ends with trailing slash before concat with additional one

SamSalvatico (approved):

apps/payments/libraries/auth.ts
alfonsograziano: Yess, just doublechecked :)

peschina (approved):
Looks good

PR #152, opened 2024-04-23
Feature: configure logging wrapper

SamSalvatico:
### Description

Write a logging wrapper for Fastify that would help us to standardize logs between services as requested in the [API Design document](https://www.notion.so/nearform/APIs-Design-Draft-bc85f57bc75e484ab1de8561f1be2794)

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**

### Checklist:

**A new PR with the test suite will follow really soon**


packages/logging-wrapper/package.json
```diff
@@ -0,0 +1,34 @@
...
+    "@types/jest": "^29.5.12",
+    "@types/node": "^20.12.7",
+    "@typescript-eslint/eslint-plugin": "^7.7.1",
+    "@typescript-eslint/parser": "^7.7.1",
+    "eslint": "^8.57.0",
+    "eslint-config-prettier": "^9.1.0",
+    "eslint-plugin-prettier": "^5.1.3",
+    "jest": "^29.7.0",
+    "prettier": "^3.2.5",
+    "ts-jest": "^29.1.2",
```
alfonsograziano, line 31: Usually, we avoid Jest while testing Node.js applications in favour of other test frameworks (like TAP) due to some [known issues](https://github.com/jestjs/jest/issues/2549).

SamSalvatico: Ouch!
I've used that because we have explicit request of do that in our [docs](https://www.notion.so/nearform/APIs-Design-Draft-bc85f57bc75e484ab1de8561f1be2794)

Anyway, thank you, I will update tests and docs!

alfonsograziano:
@SamSalvatico I like the approach! Would be cool to see some usage examples as well. Can you please add just a basic usage example? Feel free to use the `payments` app :)

packages/logging-wrapper/src/fastify-logging-wrapper.ts
```diff
@@ -0,0 +1,139 @@
...
+  };
+
+  server.setErrorHandler(function (error, request, reply) {
+    setErrorHeaders(error, reply);
+    if (!reply.statusCode || reply.statusCode === 200) {
+      const statusCode = error.statusCode ?? 500;
+      reply.code(statusCode >= 400 ? statusCode : 500);
+    }
+
+    reply.send(buildErrorResponse(error, request));
```
GiuliaTeggi, line 71: I wouldn't expect a logging package to also return errors to the client? Is there a way to keep the sending of the response as a responsibility of each api?

SamSalvatico: Thank you for your thought Giulia

TLDR; I would like to move this behaviour to another shared place, as soon as we have other utils

I agree that the logging-wrapper is not the best place to do it, but I would like to normalize the error handling across all the services to expose a single error format

Anyway, if we need to customize the error response, we can override the error handler easily after the `initializeLoggingHooks` call just invoking
```
app.setErrorHandler(function (error, request, reply) {
    reply.send({ what_we_need: true });
  });
 ```
 
 What do you think about that?

GiuliaTeggi: Yes I agree on having shared error handling, but couldn't we already setup a shared error-wrapper thing? Why do we need to wait for other utils? :) 

I just feel uneasy about mixing error handling with logging, a separate package would maintain clarity and separation of concerns.

Thurfjell: Absolutely separate concerns. 
Don't make smart things that do multiple things. Let logger log, and only log. Let http handler functions handle the requests.

Thurfjell (changes requested):

packages/logging-wrapper/src/fastify-logging-wrapper.ts
SamSalvatico: Okay all, I will create an error-handler PR on top of this PR

SamSalvatico:
@GiuliaTeggi @Thurfjell I've merged previous PRs on this, splitting concerns between `logging-wrapper` and `error-handler`

GiuliaTeggi (approved):

SamSalvatico (commented):
Readded

PR #155, opened 2024-04-23
Use stripe provider credentials

peschina:
### Ticket:

- [15503](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/15503)

### Description

Use credentials saved with Stripe provider to create a Stripe instance. If credentials are invalid we're still falling back to our account credentials set in the environment variables.

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**

### Checklist:

### Screenshots:

N/A

### DIFF
--- a/apps/payments/app/[locale]/(hosted)/paymentRequest/complete/page.tsx
+++ b/apps/payments/app/[locale]/(hosted)/paymentRequest/complete/page.tsx
@@ -1,22 +1,17 @@
-import { getTranslations } from "next-intl/server";
 import { pgpool } from "../../../../dbConnection";
 import { RedirectType, notFound, redirect } from "next/navigation";
-import {
-  getInternalStatus,
-  getPaymentIntent,
-} from "../../../../integration/stripe";
+import { getInternalStatus } from "../../../../integration/stripe";
 import { PgSessions } from "auth/sessions";
 import { TransactionStatuses } from "../../../../../types/TransactionStatuses";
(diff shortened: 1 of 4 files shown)


apps/payments/app/[locale]/(hosted)/paymentRequest/complete/page.tsx
```diff
@@ -1,22 +1,17 @@
...
-        error?: string | undefined;
-        payment_id?: string;
-        payment_intent?: string;
-      }
-    | undefined;
+  searchParams: {
+    error?: string | undefined;
+    payment_id?: string;
+    payment_intent?: string;
+    redirect_status?: string;
```
Comment (alfonsograziano) [inline apps/payments/app/[locale]/(hosted)/paymentRequest/complete/page.tsx:13]: Did you check that works with all the providers? I never noticed this `redirect_status`

alfonsograziano (approved):
LGTM - apart from a minor comment, just to doublecheck :)

apps/payments/app/[locale]/(hosted)/paymentRequest/complete/page.tsx
Comment (peschina) [reply apps/payments/app/[locale]/(hosted)/paymentRequest/complete/page.tsx:13]: It does. The `redirect_status` search params is appended by stripe only. For now we're basically just handling two uses case (stripe and truelayer), we'll refactor and extend the logic of this page when we'll add more providers

PR #162, opened 2024-04-24
Pay page layout

nnorbert:
Pay page's layout is controlled by a new query param, named `embed`.


alfonsograziano (approved):
LGTM

nnorbert:
@GiuliaTeggi @edge33 Payment page takes a new query param now, called `embed=true`. This has to be set in case of embedding the page into an iframe, so the full layout (header/footer) won't be rendered, only the content.

PR #166, opened 2024-04-25
15517 citizen payments page

nnorbert:
Citizen payments page.

### DIFF
--- a/apps/payments-api/openapi-definition.yml
+++ b/apps/payments-api/openapi-definition.yml
@@ -1240,5 +1240,103 @@ paths:
                   - error
                   - message
                   - time
+  /api/v1/citizen/transactions/{transactionId}:
+    get:
+      tags:
+        - Transactions
+      parameters:
+        - schema:
+            type: string
+          in: path
+          name: transactionId
+          required: true
+      responses:
+        "200":
+          description: Default Response
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  transactionId:
+                    type: string
+                  status:
+                    anyOf:
+                      - type: string
+                        enum:
+                          - initiated
+                      - type: string
+                        enum:
+                          - pending
+                      - type: string
+                        enum:
+                          - succeeded
+                      - type: string
+                        enum:
+                          - cancelled
+                      - type: string
+                        enum:
+                          - failed
+                  amount:
+                    type: number
+                  updatedAt:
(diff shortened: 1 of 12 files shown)


apps/payments-api/openapi-definition.yml
```diff
@@ -379,12 +325,8 @@ paths:
                 name:
                   type: string
                 data:
-                  oneOf:
-                    - $ref: '#/components/schemas/openBankingData'
-                    - $ref: '#/components/schemas/bankTransferData'
-                    - $ref: '#/components/schemas/stripeData'
-                    - $ref: '#/components/schemas/worldpayData'
-                    - $ref: '#/components/schemas/realexData'
+                  type: object
```
peschina, line 328: This change together with the one above looks like a regression, it had been introduced in a previous PR

apps/payments-api/openapi-definition.yml
```diff
@@ -856,7 +798,22 @@ paths:
                     transactionId:
                       type: string
                     status:
-                      type: string
+                      anyOf:
```
peschina, line 801: Could we add the status enum to the components schema? This way we could just reference its ref instead of duplicating it

apps/payments/app/[locale]/(hosted)/paymentSetup/providers/[providerId]/EditRealexForm.tsx
```diff
@@ -28,7 +28,7 @@ export default async ({ provider }: Props) => {
 
     await new Payments(userId).updateProvider(provider.id, {
       name: providerName,
-      data: providerData,
+      data: providerData as unknown as any, // Temp fix, Typebox type must be fixed
```
Comment (peschina) [inline apps/payments/app/[locale]/(hosted)/paymentSetup/providers/[providerId]/EditRealexForm.tsx:31]: We don't need the cast if types are updated in the openapi specification file (see comment above)

peschina (commented):
Overall looks good, there is just one issue we need to address :)

apps/payments-api/openapi-definition.yml
nnorbert: I don't know, it was not intentionally... But we must work on the types anyway so I just ignored that for the moment. Should I revert that?

peschina: As agreed in our chat we'll align the provider data at a later stage

peschina (approved):
LGTM

PR #167, opened 2024-04-25
Update profile address details

GiuliaTeggi:
### Description

- new `add-details` page to add addresses details such as ownership status and if primary residence
- move address forms steps to separate urls
- add address `ownership_status` and `is_primary` fields to profile api and FE addresses components to allow seeing, adding and editing them (these columns already existed on the `user_addresses` table). I tried to make `ownership_status` an enum but failed (had issues with openapi not picking up the fact it could be null) - if anyone wants to help that's appreciated.
- add `slug` column to `form_errors` table.
- remove useless next-inlt `navigation` components to go back to using the standard Link and redirect methods
- lock next.js version to `14.2.2` for profile

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [x] **Dev change**

### Checklist:

### Screenshots:

N/A


apps/profile/app/[locale]/addresses/add-details/[id]/page.tsx
```diff
@@ -0,0 +1,262 @@
...
+    redirect("/");
+  }
+
+  async function cancelAction() {
+    "use server";
+
+    if (!addressId) {
+      throw Error("Address id not found");
+    }
+    const { error } = await new Profile(userId).deleteAddress(addressId);
```
Comment (Thurfjell) [inline apps/profile/app/[locale]/addresses/add-details/[id]/page.tsx:98]: Does this clear any state? Seems harsh to delete stuff and throw? 
Should an implementation detail live in the public api? (I don't think so but others might disagree, I think each next app can have their own agnostic persistence for state eg.)

apps/profile/app/[locale]/addresses/edit-address/[id]/page.tsx
```diff
@@ -130,18 +148,27 @@ async function editAddress(formData: FormData) {
...
       address_line1: addressFirst,
       address_line2: addressSecond,
       town: town,
       county: county,
       eirecode: eirecode,
       move_in_date: moveInDate,
       move_out_date: moveOutDate,
+      ownership_status: isOwner === "true" ? "owner" : "renting",
+      is_primary: isPrimaryAddress === "true" ? true : false,
     });
```
Comment (Thurfjell) [inline apps/profile/app/[locale]/addresses/edit-address/[id]/page.tsx:169]: Are we ok with snake case like this? 
I think we should use either camel or snake when we now expose public apis.

apps/profile/app/[locale]/addresses/search-address/page.tsx
```diff
@@ -0,0 +1,89 @@
...
+    userId,
+    routes.addresses.searchAddress.slug,
+  );
+
+  async function searchAction(formData: FormData) {
+    "use server";
+
+    const searchQuery = formData.get("newAddress");
+
+    if (!searchQuery?.toString().length) {
```
Comment (Thurfjell) [inline apps/profile/app/[locale]/addresses/search-address/page.tsx:22]: Is it an error to prompt empty? I would imagine it resets your filtration or do nothing.

apps/profile/app/components/Addresses.tsx
```diff
@@ -42,6 +37,39 @@ export default async () => {
     }
   };
 
+  const calculateTenancyDuration = async (
+    move_in_date: Address["move_in_date"],
+    move_out_date: Address["move_out_date"],
```
Thurfjell, line 42: Why snake?

apps/profile/app/components/Addresses.tsx
```diff
@@ -42,6 +37,39 @@ export default async () => {
...
+
+    let result = "";
+    if (years > 0) {
+      result += `${years} ${translatedYears}`;
+    }
+    if (months > 0) {
+      if (years > 0) {
+        result += ", ";
+      }
+      result += `${months} ${translatedMonths}`;
```
Thurfjell, line 67: You could probably also do an array and join ", " if you want?

apps/profile/migrations/sqls/20240422103826-update-form-errors-table-up.sql
```diff
@@ -0,0 +1,5 @@
+BEGIN; 
+
+ALTER TABLE form_errors ADD COLUMN slug TEXT NOT NULL;
+
+COMMIT;
```
Thurfjell, line 5: Uh oh is this how you're supposed to dow with begin/commit? 😲

Thurfjell (commented):
Can't we just use camel everywhere in javascript? 😢

edge33 (approved):
LGTM

apps/profile/app/[locale]/addresses/add-details/[id]/page.tsx
Comment (GiuliaTeggi) [reply apps/profile/app/[locale]/addresses/add-details/[id]/page.tsx:98]: I don't fully follow you here, I am deleting the address that was added the step before the add-details one when a user clicks on cancel because it would be weird if they cancelled adding an address but still saw it appearing in the homepage? Let's catch up on slack.

apps/profile-api/routes/addresses/index.ts
```diff
@@ -149,35 +149,23 @@ export default async function addresses(app: FastifyInstance) {
...
-        address_line1,
-        address_line2,
-        town,
-        county,
-        eirecode,
-        move_in_date,
-        move_out_date,
-      } = request.body;
+
+      const keys = Object.keys(request.body);
```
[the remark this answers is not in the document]
GiuliaTeggi: Yeah, ended up doing this given that any field is optional, but will look into better ways to handle this.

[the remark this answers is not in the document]
GiuliaTeggi: The schema will actually check which fields are being sent and error if any fields are not expected, so this should be fine

apps/profile/app/[locale]/addresses/add-details/[id]/page.tsx
Comment (Thurfjell) [inline apps/profile/app/[locale]/addresses/add-details/[id]/page.tsx:1]: Hmm this resource url looks strange to me. 
What is the id? Address id? 
Then it should probably be addresses/[id[/details no?

apps/profile-api/routes/addresses/index.ts
```diff
@@ -149,35 +168,94 @@ export default async function addresses(app: FastifyInstance) {
...
+
+        if (!result?.rows.length) {
+          const error = app.httpErrors.notFound("Address not found");
+          error.statusCode = 404;
+          error.code = "NOT_FOUND";
+
+          throw error;
+        }
+
+        reply.send({ id: result.rows[0].id });
```
Thurfjell, line 211: Not sure what standards are or if we want to use any, but could probably return number of rows updated instead of id.

apps/profile-api/types/schemaDefinitions.ts
```diff
@@ -37,9 +39,27 @@ export const ParamsWithAddressId = Type.Object({
...
+  isPrimary: Type.Boolean(),
+  ownershipStatus: Type.String(),
+});
+
 export type UpdateAddress = Static<typeof UpdateAddress>;
 
+export const PatchAddress = Type.Object({
+  isPrimary: Type.Optional(Type.Boolean()),
+  ownershipStatus: Type.Optional(Type.String()),
+});
```
Thurfjell, line 59: I guess it's extendable and this is the only thing we need, and whenever something else needs updating, it's added here but I just guess I wouldn't mind if all fields were available.

apps/profile/app/[locale]/addresses/add-details/[id]/page.tsx
Comment (GiuliaTeggi) [reply apps/profile/app/[locale]/addresses/add-details/[id]/page.tsx:1]: Yes you're right

apps/profile-api/types/schemaDefinitions.ts
GiuliaTeggi: What made me not add all of the fields is that most of the other fields cannot be set to null, I could add the other two fields that are optional here, but at the moment only these two need to be patched.

apps/profile-api/routes/addresses/index.ts
GiuliaTeggi: I think either returning the id or the number of rows updated is the same in this case, given there's only one address to update? In `payments-api` row id is used, so I kept it the same.

apps/profile/app/[locale]/addresses/add-details/[id]/page.tsx
Comment (GiuliaTeggi) [reply apps/profile/app/[locale]/addresses/add-details/[id]/page.tsx:1]: Updated

apps/profile-api/routes/addresses/index.ts
GiuliaTeggi: So, after speaking with @SamSalvatico, it looks like the OGCIO api design document is only explicit about returning the id on successful `POST` requests, but it doesn't prescribe anything re what needs to be returned on successful `PUT` or `PATCH`...

Thurfjell: Yes id on post is all fine.

For put and patch I guess we could argue a NoContent is fine.

GiuliaTeggi: I'll open a discussion on the team channel if this is blocking for you, just to be all on the same page given it's not something we discussed before.

PR #171, opened 2024-04-26
Feature/error handling

SamSalvatico:
### Description

Added an `error-handling` package to normalize handling across packages

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**

### Checklist:


apps/payments-api/app.ts
```diff
@@ -14,6 +14,7 @@ import { dirname, join } from "path";
 import healthCheck from "./routes/healthcheck";
 import sensible from "@fastify/sensible";
 import schemaValidators from "./routes/schemas/validations";
+import { initializeErrorHandling } from "error-handling";
```
GiuliaTeggi, line 17: Super nitpick, but I would call this error-handlER (to follow the same pattern of logging-wrappER)

GiuliaTeggi (approved):
LGTM 😄  Great job on these utils!!

Thurfjell (commented):
Q out of ignorance and not understanding. 
What, how and where are we logging in ie. prod?

apps/payments-api/app.ts
SamSalvatico: You're totally right, thank you Giulia!

SamSalvatico:
> Q out of ignorance and not understanding. What, how and where are we logging in ie. prod?

We will log same entries in whole environments. The only thing that will change is the minimum logged level.
e.g. we can decide to not log `DEBUG` level on prod

https://github.com/ogcio/life-events/blob/feature/logging-wrapper-tests/packages/logging-wrapper/README.md#default-records

Thurfjell (approved):

PR #179, opened 2024-04-29
Add mock Irish addresses

GiuliaTeggi:
### Description

Return mock Irish addresses from addresses mock api instead of `faker.js` ones. I have just commented out the logic that was using the search key, so to always return some results (with faker we were generating 5000 addresses and running the search on those). 
## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [x] **Dev change**

### Checklist:

### Screenshots:

N/A


edge33 (approved):
LGTM

PR #182, opened 2024-04-29
feat(timeline-api) Added timeline api

edge33:
### Description

This change adds the timeline-api to the apps

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**


apps/timeline-api/app.ts
```diff
@@ -0,0 +1,105 @@
...
+      tags: [
+        {
+          name: "User",
+        },
+        {
+          name: "Addresses",
+        },
+        {
+          name: "Entitlements",
+        },
```
GiuliaTeggi, line 46: These tags are not relevant for the timeline - you should set different ones up if/when needed. have you checked this on swagger?

apps/timeline-api/app.ts
```diff
@@ -0,0 +1,105 @@
...
+   * postgres not required right now, leaving it commented for now
+   */
+  // app.register(postgres, {
+  //   host: process.env.POSTGRES_HOST,
+  //   port: Number(process.env.POSTGRES_PORT),
+  //   user: process.env.POSTGRES_USER,
+  //   password: process.env.POSTGRES_PASSWORD,
+  //   database: process.env.POSTGRES_DB_NAME,
+  // });
+
```
GiuliaTeggi, line 72: I think you should uncomment this so that when we merge this pr we can check that all is hooked up correctly with the env vars - even if the db is not used right now. @williamwd should be notified too, so he can setup the required variables in the dev environment

apps/timeline-api/index.ts
```diff
@@ -0,0 +1,16 @@
+import { writeFile } from "fs/promises";
+
+import { build } from "./app";
+
+const app = await build({ logger: true });
+
+app.listen({ host: "0.0.0.0", port: 8003 }, (err, address) => {
```
GiuliaTeggi, line 7: This is the profile api port, please find another one that is not used by any other app :D

apps/web/app/[locale]/[event]/components/Timeline.tsx
```diff
@@ -45,12 +44,6 @@ export default async ({ searchProps, messages, locale }: TimelineProps) => {
 
   queryParams.set("searchQuery", searchQuery);
 
-  const submitQuery = submitQuery_.bind(
-    null,
-    path,
-    new URLSearchParams(searchProps),
-  );
-
```
Comment (GiuliaTeggi) [inline apps/web/app/[locale]/[event]/components/Timeline.tsx:53]: Just double checking this was no longer needed?

Comment (edge33) [reply apps/web/app/[locale]/[event]/components/Timeline.tsx:53]: it was unused, so I removed it

apps/web/app/[locale]/timeline/page.tsx
```diff
@@ -1,10 +1,10 @@
 import { PgSessions } from "auth/sessions";
 import { web } from "../../utils";
-import Timeline, { TimeLineData } from "./Timeline";
-import getTimelineData from "../../data/getTimelineData";
+import Timeline from "./Timeline";
+import { Timeline as TimelineClient } from "building-blocks-sdk";
```
Comment (GiuliaTeggi) [inline apps/web/app/[locale]/timeline/page.tsx:4]: If you want to alias this, can you do it everywhere the Timeline client is used?

apps/web/app/[locale]/[event]/components/Timeline.tsx
```diff
@@ -79,7 +79,7 @@ export default async ({ searchProps, messages, locale }: TimelineProps) => {
       >
         <div style={{ borderLeft: "1px solid #B1B4B6", paddingLeft: "10px" }}>
           {timelineData?.data &&
-            timelineData.data.reverse().map((yearObject) => {
+            timelineData.data.data.reverse().map((yearObject) => {
```
Comment (GiuliaTeggi) [inline apps/web/app/[locale]/[event]/components/Timeline.tsx:82]: I think not to have to call data.data all the times you could format the query result in the SDK so that data is not further nested?

Comment (edge33) [reply apps/web/app/[locale]/[event]/components/Timeline.tsx:82]: in this casa data is an actual property of the TimeLine response object, which is defined as

```
export type TimeLineData = {
  minYear: number;
  maxYear: number;
  data: {
    year: number;
    minYear: number;
    maxYear: number;
    months: Month[];
  }[];
};
```

Comment (GiuliaTeggi) [reply apps/web/app/[locale]/[event]/components/Timeline.tsx:82]: ha I see, ok then :)

GiuliaTeggi (approved):

edge33:
@williamwd we would need the following .env vars for this service

POSTGRES_USER
POSTGRES_PASSWORD
POSTGRES_HOST
POSTGRES_PORT
POSTGRES_DB_NAME=timeline
POSTGRES_DB_NAME_SHARED=shared

apps/timeline-api/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+COPY ./apps/timeline-api/ /app/apps/timeline-api/
+
+FROM deps AS builder
+WORKDIR /app
+
+ENV NODE_ENV=production
+ENV LOG_LEVEL=trace
+
+EXPOSE 8003
```
williamwd, line 18: Please consider using 8004, as `profile-api` already uses 8003.

edge33: nice catch, thanks mate

williamwd (approved):

PR #183, opened 2024-04-30
Realex integration

peschina:
### Ticket:

- [4545](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/4545)

### Description

Realex provider integration. Flow:
- user chooses Realex as payment method
- api is called to retrieve payment object
- user is redirected to new page. An hidden form is submitted when page loads - this triggers the HPP flow.
- user is redirected to Realex HPP and fills in all required info to make the payment
- user clicks to proceed with the payment and Realex does a POST api call to our endpoint with transaction info. We ensure the call is legit by checking the hash provided by them. If all is good we return an HTML
- The HTML is rendered in the page (the user is still in the page hosted by Realex). Nothing is actually rendered, the user is redirected back to our application where we resume our usual flow
- The transaction status is updated and the user is redirected to the url specified in the payment request

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**

### Checklist:

### Screenshots:

https://github.com/ogcio/life-events/assets/47299026/ca765731-059e-45ca-9d35-01064a70008f


apps/payments/app/[locale]/(hosted)/paymentRequest/realex/RealexHost.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+const PAYMENT_DEFAULT_COUNTRY = "IE";
+
+export default function RealexHost({ payment, locale, responseUrl }: Props) {
+  const t = useTranslations();
+
+  const formRef = useRef(null);
+
+  useEffect(() => {
+    formRef.current && (formRef.current as HTMLFormElement).submit();
```
Comment (peschina) [inline apps/payments/app/[locale]/(hosted)/paymentRequest/realex/RealexHost.tsx:19]: By submitting the form we initiate the hpp flow

apps/payments-api/routes/transactions/index.ts
```diff
@@ -245,4 +251,162 @@ export default async function transactions(app: FastifyInstance) {
       reply.send(result.rows[0]);
     },
   );
+
+  app.get<{
```
peschina, line 255: We call this to get the payment object that we need to feed to Realex HPP. It contains the realex account keys and an hash created with the secret key and the payment sensitive data

apps/payments-api/routes/transactions/index.ts
```diff
@@ -245,4 +251,162 @@ export default async function transactions(app: FastifyInstance) {
...
+
+      const providerSecretsHandler = providerSecretsHandlersFactory("realex");
+      const { merchantId, sharedSecret } =
+        providerSecretsHandler.getClearTextData(
+          provider.data,
+        ) as unknown as RealexData;
+
+      const currency = "EUR";
+      const url =
+        process.env.REALEX_PAYMENT_URL ??
```
peschina, line 309: For now I think it's fine if we default to sandbox url - let me know if you prefer to throw an error if the env var is missing

apps/payments-api/routes/transactions/index.ts
```diff
@@ -245,4 +251,162 @@ export default async function transactions(app: FastifyInstance) {
...
+        throw app.httpErrors.unprocessableEntity("Provider is not enabled");
+      }
+
+      const providerSecretsHandler = providerSecretsHandlersFactory("realex");
+      const { merchantId, sharedSecret } =
+        providerSecretsHandler.getClearTextData(
+          provider.data,
+        ) as unknown as RealexData;
+
+      const currency = "EUR";
```
peschina, line 307: At some point we'll have to discuss if we want to support euro currency only and how to deal with it consequently

apps/payments-api/routes/transactions/index.ts
```diff
@@ -245,4 +251,162 @@ export default async function transactions(app: FastifyInstance) {
...
+      const isResponseValid = realexService.verifyHash(body);
+      if (!isResponseValid) {
+        throw app.httpErrors.unprocessableEntity(
+          "Payment response contains untrusted data",
+        );
+      }
+
+      const res = realexService.generateHTMLResponse(body);
+
+      reply.header("Content-Type", "text/html");
```
peschina, line 408: HPP will render the HTML response received when calling this endpoint. The redirect is on us.

apps/payments-api/routes/transactions/index.ts
```diff
@@ -245,4 +251,162 @@ export default async function transactions(app: FastifyInstance) {
...
+        process.env.REALEX_PAYMENT_URL ??
+        "https://pay.sandbox.realexpayments.com/pay";
+      const realexService = new RealexService(sharedSecret);
+      const timestamp = realexService.generateTimestamp();
+      const hash = realexService.generateHash(
+        [timestamp, merchantId, intentId, amount, currency].join("."),
+      );
+
+      const result = {
+        ACCOUNT: process.env.REALEX_PAYMENT_ACCOUNT ?? "internet",
```
peschina, line 318: If this is a sub-account it should provided by the user when creating the provider. If confirmed I'll address this in a separate PR

apps/payments-api/routes/transactions/index.ts
```diff
@@ -245,4 +251,162 @@ export default async function transactions(app: FastifyInstance) {
...
+        throw app.httpErrors.unprocessableEntity("Provider is not enabled");
+      }
+
+      const providerSecretsHandler = providerSecretsHandlersFactory("realex");
+      const { sharedSecret } = providerSecretsHandler.getClearTextData(
+        provider.data,
+      ) as unknown as RealexData;
+
+      const realexService = new RealexService(sharedSecret);
+      const isResponseValid = realexService.verifyHash(body);
```
peschina, line 399: The hash contained in the response is created by Realex with the shared secret. If we can create the same hash it means that the request is legit

apps/payments-api/routes/transactions/index.ts
alfonsograziano: This concern is valid for all payment providers. We will put this in the roadmap at some point in case will be needed :)

apps/payments-api/routes/transactions/index.ts
alfonsograziano, line 255: I know that we're putting everything under the same file, but once the file starts to be over 400 lines maybe is worth splitting it? Shall we have a `realex.ts` inside the `transactions` folder with any provider-specific API?

apps/payments-api/routes/transactions/index.ts
alfonsograziano: I'm just concerned that we forget (maybe 6 months from now) to add that var and then suddenly we use a wrong url instead of throwing an error (which is much more visible). 

In case that's possible (maybe in a separate PR) we should have a validation with all the necessary env variables. What do you think? If a variable is required, we should throw an error on the startup

apps/payments-api/routes/transactions/index.ts
alfonsograziano: non-blocking: Shall we add a comment here explaining the code :) I feel every time we need to add a comment on the PR, that comment would probably be useful also in the code as well for future reference

apps/payments-api/services/cryptographyService.ts
```diff
@@ -1,21 +1,27 @@
 import crypto from "crypto";
 
 class CryptographyService {
-  private algorithm: string;
-  private key: Buffer;
+  private encryptionAlgorithm: string;
```
alfonsograziano, line 4: non-blocking: I'd personally create another Service specialized in hashing rather than overcharging the current service (even because it contains sensitive business logic), but also your choice is fine :)

apps/payments-api/routes/transactions/index.ts
peschina: I agree. Let's address this separately

apps/payments-api/routes/transactions/index.ts
peschina: Sure. At some point we'll also talk about eventually moving the logic from the route handlers

apps/payments-api/routes/transactions/index.ts
peschina: [3744ee8](https://github.com/ogcio/life-events/pull/183/commits/3744ee89f0fb5ef423ee3197bbb08a9fe36603d3)

apps/payments-api/routes/transactions/index.ts
```diff
@@ -245,4 +251,162 @@ export default async function transactions(app: FastifyInstance) {
...
+
+      const provider = providerRes.rows[0];
+      if (provider.status !== "connected") {
+        throw app.httpErrors.unprocessableEntity("Provider is not enabled");
+      }
+
+      const providerSecretsHandler = providerSecretsHandlersFactory("realex");
+      const { sharedSecret } = providerSecretsHandler.getClearTextData(
+        provider.data,
+      ) as unknown as RealexData;
```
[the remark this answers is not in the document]
peschina: Good catch. We didn't have the correct types when I first implemented the service, let me amend that

apps/payments-api/routes/transactions/index.ts
```diff
@@ -245,4 +251,162 @@ export default async function transactions(app: FastifyInstance) {
...
+          422: HttpError,
+        },
+      },
+    },
+    async (request, reply) => {
+      const body = request.body;
+
+      let transaction;
+      try {
+        transaction = await app.pg.query(
```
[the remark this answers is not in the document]
peschina: Good catch, I didn't circle back after it was merged. Let's clean up the other handlers as well in a separate PR.
[654acc7](https://github.com/ogcio/life-events/pull/183/commits/654acc7ca48f3d002d93be11867fc14c70bab89c)

apps/payments-api/services/cryptographyService.ts
peschina: I don't particularly mind one choice or another, tbh I didn't bother because we'll remove all the encryption logic as soon as we have the secret manager

apps/payments-api/routes/transactions/index.ts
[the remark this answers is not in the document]
peschina: [afece04](https://github.com/ogcio/life-events/pull/183/commits/afece04cb860abaec580b4253583aa11e91a877c)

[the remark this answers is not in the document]
peschina: Actually, this is a broader impact so let's deal with it separately

apps/payments-api/routes/transactions/index.ts
peschina: [d3d029f](https://github.com/ogcio/life-events/pull/183/commits/d3d029f98f129f87a921380954d4a85ee9edef88)

PR #185, opened 2024-04-30
Setup digital wallet onboarding

GiuliaTeggi:
### Description

Adding digital wallet onboarding steps with gov id not verified and verified flows. 

Placed timeline behind a feature flag (`timeline`) to hide it and add new feature flag to show the digital wallet onboarding (`digitalWallet`).

⚠️ feature flags will need to be seeded into the db before the launch.

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**

### Checklist:

### Screenshots:

N/A


apps/web/app/[locale]/[event]/MyLifeEvents.tsx
```diff
@@ -9,53 +9,95 @@ import { orderBirthCertificateRules } from "./[...action]/OrderBirthCertificate/
...
+    default:
+      throw new Error(`Unsupported workflow: ${flow}`);
+  }
 };
 
 async function getEvents() {
   "use server";
 
   return Promise.resolve([
+    // All the commented out events are now reachable via the burger menu
```
Comment (edge33) [inline apps/web/app/[locale]/[event]/MyLifeEvents.tsx:57]: should we delete this?

apps/web/app/[locale]/[event]/[...action]/GetDigitalWallet/ChangeDetails.tsx
```diff
@@ -0,0 +1,322 @@
...
+
+  const red = ds.colours.ogcio.red;
+
+  const errors = await form.getErrorsQuery(
+    userId,
+    routes.digitalWallet.getDigitalWallet.changeDetails.slug,
+    flow,
+  );
+
+  async function submitAction(formData: FormData) {
```
Comment (edge33) [inline apps/web/app/[locale]/[event]/[...action]/GetDigitalWallet/ChangeDetails.tsx:25]: non blocking, nit: I believe we should standardise the way we scaffold server action for better readability, decide whether we want to adopt guidelines for separate folder/dedicated path

edge33 (approved):
LGTM

apps/web/app/[locale]/[event]/[...action]/GetDigitalWallet/ChangeDetails.tsx
Comment (GiuliaTeggi) [reply apps/web/app/[locale]/[event]/[...action]/GetDigitalWallet/ChangeDetails.tsx:25]: Sure, at the moment this is the way all form submit actions are setup for life events workflows. I don't have anything against anyone refactoring this at any point, it probably should be refactored as you say :)

PR #196, opened 2024-05-03
Payment request basic status management

alfonsograziano:
### Ticket:

- [15434](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/15434)

### Description

The Payment request has now a "active" and "inactive" status

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [X] **New feature**
- [ ] **Dev change**

### Screenshots:
Screenshot 2024-05-03 at 14 07 54
Screenshot 2024-05-03 at 14 07 49


apps/payments-api/routes/schemas/index.ts
```diff
@@ -157,6 +157,7 @@ export const PaymentRequest = Type.Object({
   amount: Type.Number(),
   reference: Type.String(),
   providers: Type.Array(ProviderDetails),
+  status: Type.Union([Type.Literal("active"), Type.Literal("inactive")]),
```
peschina, line 160: Can you create a type for this status and then use it here and also in the paymentRequests handlers? Otherwise status will be an enum in some places and just a string in other places

peschina: Also, do we think we'll add other possible states? Or should status be just a boolean `active`?

apps/payments/app/[locale]/(hosted)/paymentSetup/create/page.tsx
```diff
@@ -23,6 +23,7 @@ async function createPayment(userId: string, formData: FormData) {
     redirectUrl: formData.get("redirect-url") as string,
     allowAmountOverride: formData.get("allowAmountOverride") === "on",
     allowCustomAmount: formData.get("allowCustomAmount") === "on",
+    status: formData.get("status") as string,
```
Comment (peschina) [inline apps/payments/app/[locale]/(hosted)/paymentSetup/create/page.tsx:26]: Can we create a type for status that reflects the way it is typed in the api? Same below

apps/payments/app/[locale]/(hosted)/paymentSetup/PaymentSetupForm.tsx
Comment (peschina) [inline apps/payments/app/[locale]/(hosted)/paymentSetup/PaymentSetupForm.tsx:1]: Looks good for now, but we might want to review the ui in the future. I would expect the request status to be presented as the other fields in the form, it feels like a different section in the page instead (with the header and the amount of space it takes)

peschina (commented):
A few comments, looks good overall

apps/payments-api/routes/schemas/index.ts
alfonsograziano: Yes, it will be possible in the future to have multiple states. We will evaluate in the next iterations what other states are required

apps/payments/app/[locale]/(hosted)/paymentSetup/create/page.tsx
Comment (alfonsograziano) [reply apps/payments/app/[locale]/(hosted)/paymentSetup/create/page.tsx]: Done, thanks :)

peschina (approved):
LGTM, just fix the build

PR #206, opened 2024-05-06
Enforce payments-api environment variables

peschina:
### Description

Throw error at startup if an environment variable for `payments-api` is missing

## Type

- [ ] **Dependency upgrade**
- [x] **Bug fix**
- [ ] **New feature**
- [ ] **Dev change**

### Checklist:

### Screenshots:

N/A

### DIFF
--- a/apps/payments-api/config.ts
+++ b/apps/payments-api/config.ts
@@ -7,6 +7,11 @@ export const envSchema = {
     "POSTGRES_PORT",
     "POSTGRES_DB_NAME",
     "POSTGRES_DB_NAME_SHARED",
+    "PAYMENT_INTENTID_LENGTH",
+    "PAYMENT_INTENTID_MAX_TRY_GENERATION",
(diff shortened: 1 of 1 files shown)


apps/payments-api/config.ts
```diff
@@ -27,5 +32,20 @@ export const envSchema = {
     POSTGRES_DB_NAME_SHARED: {
       type: "string",
     },
+    PAYMENT_INTENTID_LENGTH: {
```
alfonsograziano, line 35: Non-blocking: do we want to do the same for Stripe?

alfonsograziano (approved):

apps/payments-api/config.ts
peschina: The stripe env variables are not in the `payments-api` project. I'd rather not enforce them as they are there temporarily and currently set as optional - the integration works fine without them

PR #214, opened 2024-05-08
Update profile app responsive styles

GiuliaTeggi:
### Description

Making profile app header and main body responsive.

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [ ] **New feature**
- [ ] **Dev change**

### Checklist:

### Screenshots:

N/A


apps/profile/app/styles/globals.scss
```diff
@@ -0,0 +1,27 @@
+$breakpoint-phone: 767px;
```
GiuliaTeggi, line 1: These breakpoints should be moved somewhere shared at some point..maybe as part of the design system.

Thurfjell (approved):

PR #258, opened 2024-05-15
feat(AuthService) Added authservice to all apps

edge33:
This change adds authservice to all apps


apps/auth-service/src/routes/static/redirect.html
```diff
@@ -94,5 +94,4 @@
 <script type="module">
   setTimeout(() => {
     document.querySelector('#redirect-form').submit()
-  }, 5000)
-</script>
+  }, %REDIRECT_TIMEOUT%)</script>
```
alfonsograziano, line 97: Any specific reason to make the redirect timeout an env variable instead of a static value? Are we planning to change the value of the timeout across multiple environments?

apps/profile/app/[locale]/addresses/[id]/add-details/page.tsx
```diff
@@ -1,4 +1,4 @@
-import { AuthServicePgSessions } from "auth/sessions";
+import { PgSessions } from "auth/sessions";
```
Comment (alfonsograziano) [inline apps/profile/app/[locale]/addresses/[id]/add-details/page.tsx:1]: Not a blocker, but more a question of the long-term. Are we planning to keep the Sessions management under the `PgSession` name or do we want to rename it to something more generic? (again, no personal preference here)

apps/auth-service/src/routes/static/redirect.html
edge33: not specifcally, but we can do some tests and check to find a value that makes sense, with no need to rebuild, once we are happy with a value, I will remove it

apps/profile/app/[locale]/addresses/[id]/add-details/page.tsx
Comment (edge33) [reply apps/profile/app/[locale]/addresses/[id]/add-details/page.tsx:1]: No Preference either, I just adapted to what i was before I started this change

apps/auth-service/src/routes/auth/utils/cookies.ts
```diff
@@ -31,12 +28,10 @@ const deleteCookie = (
...
   key: string,
   value: string,
-  appConfig: { [key: string]: string },
 ) => {
   const secure = isSecure(request);
   reply.cookie(key, value, {
     path: "/",
-    // domain: appConfig.HOST_DOMAIN,
     httpOnly: true,
     expires: new Date(Date.now() - 100),
```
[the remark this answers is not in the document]
SamSalvatico: oh, got it, thanks

package.json
```diff
@@ -14,6 +14,7 @@
     "build:auth": "npm run build --workspace=auth",
     "build:logging-wrapper": "npm run build --workspace=logging-wrapper",
     "build:error-handler": "npm run build --workspace=error-handler",
+    "build:services": "npm run build --workspace=auth-service && npm run build --workspace=web && npm run build --workspace=payments && npm run build --workspace=messaging && npm run build --workspace=profile && npm run build --workspace=timeline-api",
```
alfonsograziano, line 17: Do we need this script?

alfonsograziano (approved):
Apart from a couple of optional comments, LGTM

package.json
edge33: I used as a utility, I will remove it

alfonsograziano: +1 on the makefile

edge33: we can add it back later when we will have the services with build scripts, for now, most of them don't have it

PR #260, opened 2024-05-15
[Payments] Responsive header

peschina:
### Ticket:

- [16652](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/16652)

### Description

Moved `Header` component to components folder and updated it to make it responsive. An hamburger button is available on mobile and opens a side menu. This is currently available with js enabled only.
Header used in info route will be changed to use the common header in a separate pr. Responsiveness of the recently added building blocks tiles will be addressed separately as well

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**

### Checklist:

### Screenshots:

Public Servant:
localhost_3001_en_paymentSetup(iPhone 12 Pro)
localhost_3001_en_paymentSetup(iPhone 12 Pro) (1)

Citizen:
localhost_3001_en_citizen_transactions(iPhone 12 Pro)


apps/payments/app/components/HamburgerMenu/HamburgerMenu.tsx
```diff
@@ -0,0 +1,132 @@
...
+          <Link
+            className={`govie-link govie-link--no-underline`}
+            style={{
+              fontWeight: locale === "en" ? "bold" : "normal",
+              color: ds.colours.ogcio.darkGreen,
+            }}
+            href={"/en/" + pathSlice}
+            prefetch={false}
+          >
+            English
```
peschina, line 107: Should the language names be translated as well?

alfonsograziano: We can keep it like this for now, not sure if we will ever need to translate it :)

apps/payments/app/components/Header/LanguageSwitch.tsx
```diff
@@ -0,0 +1,36 @@
+import ds from "design-system";
+import { headers } from "next/headers";
+
+export default () => {
+  const pathSlice = headers().get("x-pathname")?.split("/") ?? [];
+  const path = pathSlice.slice(2)?.join("/") || "";
+
+  return (
+    <div style={{ display: "flex", alignItems: "center", gap: "10px" }}>
```
alfonsograziano, line 9: Is this code duplication? It looks to me that we have the language switch into two places (here and in `HamburgerMenu.tsx`)

alfonsograziano (approved):
Apart from a component that might need refactoring, LGTM

apps/payments/app/components/HamburgerMenu/HamburgerMenu.module.scss
```diff
@@ -0,0 +1,90 @@
+@import "/app/styles/globals.scss";
```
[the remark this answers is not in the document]
peschina: For sure. Our implementation is very similar to the one in web, but not quite the same. A bit simpler and leaner I hope :)

PR #266, opened 2024-05-16
Api Auth plugin

alfonsograziano:
### Description

A simple auth plugin for Fastify, based on auth-service

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [X] **New feature**
- [ ] **Dev change**


apps/auth-service/src/routes/session/index.ts
```diff
@@ -0,0 +1,46 @@
...
+import { Type, Static } from "@sinclair/typebox";
+
+const QueryStringSchema = Type.Object({
+  sessionId: Type.String(),
+});
+
+type QueryString = Static<typeof QueryStringSchema>;
+
+export default async (app: FastifyInstance) => {
+  app.get(
```
edge33, line 11: curiosity: since we are willing to call this from the APIs why GET and not POST?

edge33 (approved):
Looks perfect, left a small comment out of curiosity

apps/auth-service/src/routes/session/index.ts
alfonsograziano: No specific reason. Usually POST requests are used to create resources, in this case we want to GET something based on something else (that's why the query param, but it can even be an header). 

I think in this specific scenario, there are no limitations, I just used the simpler solution I could think of :)

SamSalvatico (approved):
LGTM

PR #273, opened 2024-05-17
Chore(auth) Authentication integration doc

edge33:
image


alfonsograziano (approved):

documentation/docs/authentication.md
```diff
@@ -29,9 +29,22 @@ sequenceDiagram
...
-To implement authentication in a NextJS app, you can use the `auth` package and call `PgSessions.get()`. This function will handle retrieving session data using the `sessionID` cookie or redirecting to the `AuthService` specified by the `AUTH_SERVICE_URL` environment variable.
+## Integrating Auth service in your webapp
 
-To log out, simply make a GET request to `/auth/logout`. Make sure to include the `redirectUrl` query parameter to specify where users should be redirected after logging out and potentially logging back in.
+Auth service checklist integration:
+
+- [ ] ENV VARS: HOST_URL set to the current application URL and AUTH_SERVICE_URL pointing to the auth service
+- [ ] Expose an API route for logging in /app/api/auth/route.ts
+
+Ideally this file should contain
```
Thurfjell, line 39: Why?

edge33: ah, it looks like I fell asleep on the keyboard 👀

edge33: - Env vars are needed in order to allow redirection to auth service and back to the web app originating the request
- the api route is needed to allow actual auth against the shared db and set the cookie from the originating request. this was done to overcome the limit of not having the possibility to allow the apps from the same domain, hence we need a cookie for every webapp
the api route file can import the login route from the auth shared package.

let me know if you need further clarification :)

PR #275, opened 2024-05-17
[Payments] Reuse language switch component in header

peschina:
### Ticket:

- [16652](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/16652)

### Description

Reuse language switch component in header and hamburger menu

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [ ] **New feature**
- [x] **Dev change**

### Checklist:

### Screenshots:

N/A


apps/payments/app/components/HamburgerMenu/index.tsx
```diff
@@ -14,7 +15,11 @@ export default async ({ userName, locale, publicServant }: HamburgerProps) => {
 
   return (
     <NextIntlClientProvider messages={menuMessages}>
-      <HamburgerMenuWrapper userName={userName} publicServant={publicServant} />
+      <HamburgerMenuProvider
+        userName={userName}
+        publicServant={publicServant}
+        languageSwitch={<LanguageSwitch theme={"light"} />}
```
alfonsograziano, line 21: I'm ok with this implementation, but just out of curiosity, is there any specific reason to not put directly `<LanguageSwitch theme="dark" />` inside the `HamburgerMenuProvider` component instead of injecting it as a prop?

alfonsograziano (approved):
LGTM

apps/payments/app/components/HamburgerMenu/index.tsx
peschina: Any child of client component is a client component too, so importing a server component in a client component is forbidden. You can pass it down as a prop or render it in a slot though. [Here](https://nextjs.org/docs/app/building-your-application/rendering/composition-patterns#interleaving-server-and-client-components) is a related page in the docs if you're curious :)

PR #280, opened 2024-05-20
Remove deprecated messages table trail in migrations

Thurfjell:
### Ticket:

- [16950](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/16950)

### Description

Remove the deprecated messages table migrations since there are three of them. 

Only the latest is relevant, thus we don't need a trail of conceptual tables with dropping in the up.

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [ ] **New feature**
- [ ] **Dev change**

### Checklist:

### Screenshots:

N/A


SamSalvatico (approved):
Tested also with a fresh environment, it works

LGTM

PR #281, opened 2024-05-20
Scheduler readme documentation

Thurfjell:
### Ticket:

- [16951](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/16951)

### Description

Add a description what the scheduler api does and some examples how to use it.

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [ ] **New feature**
- [ ] **Dev change**

### Checklist:

### Screenshots:

N/A


apps/scheduler-api/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+Recall same endpoint to simulate a cron like behaviour
+
+```javascript
+someApp.Post(
+  "your.client.handler/jobs/:id",
+  async function jobHandler(req, res) {
+    const id = req.params.id;
+    await logic(id);
+    fetch(...same original call with execute at now() + arbitrary unit of time)
```
peschina, line 74: I'm not sure whether this pseudo code is better or worse than just a comment. Maybe worth seeing what the others think as well

peschina (approved):
LGTM!

apps/scheduler-api/README.md
edge33: I would prefer to have sample with the actual fetch url.

apps/scheduler-api/README.md
```diff
@@ -0,0 +1,108 @@
...
+someApp.Post(
+  "your.client.handler/jobs/:id",
+  async function jobHandler(req, res) {
+    const id = req.params.id;
+    void doSomethingReallyLengthy(id);
+  },
+);
+```
+
+Recall same endpoint after a day, to simulate a cron like behaviour
```
Thurfjell, line 66:
```suggestion
Recall same endpoint after an interval, to simulate a cron like behaviour
```

edge33:
LGTM

edge33 (approved):

PR #282, opened 2024-05-20
Feature: models for messaging users

SamSalvatico:
### Description

First migrations and models to make organization users to import users.
Logic and APIs will be created in next PRs

Flow explained:

- a user from an organization will be able to import users in two ways,  through `csv` files or through `api`
- once the organisation user has started the import, the data will be saved in the `users_imports` table to be imported
- we can then decide if start the process in a sync or async way
- the process will read from the`users_import` table, then process the users in the json one by one, trying to map them through queries to the user-profile APIs
    - if it exists, will be set as imported, added to the users table if needed and to the organisations_users table
    - if it doesn't, will be set as not_found
- once all the users from an import have been imported, the import entry will be deleted
- after X retries, we won't retry anymore

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**


packages/messages/migrations/sqls/20240520080749-add-users-data-up.sql
```diff
@@ -0,0 +1,33 @@
...
+    users (
+        id uuid NOT NULL DEFAULT gen_random_uuid () PRIMARY KEY,
+        user_profile_id uuid,
+        importer_organisation_id uuid NOT NULL,
+        user_status text NOT NULL DEFAULT 'pending',
+        -- used to describe the quality 
+        -- of the correlation between user profile and the imported users
+        -- e.g. if the mapping is made by PPSN is full, 
+        -- if thanks to name and birthday, partial
+        correlation_quality text NOT NULL DEFAULT 'full'
```
Thurfjell, line 11: What exactly is this used for?

packages/messages/migrations/sqls/20240520080749-add-users-data-up.sql
```diff
@@ -0,0 +1,33 @@
+CREATE TABLE
+    users (
+        id uuid NOT NULL DEFAULT gen_random_uuid () PRIMARY KEY,
+        user_profile_id uuid,
+        importer_organisation_id uuid NOT NULL,
+        user_status text NOT NULL DEFAULT 'pending',
```
Thurfjell, line 6: This is some kind of on boarding status?

packages/messages/migrations/sqls/20240520080749-add-users-data-up.sql
```diff
@@ -0,0 +1,33 @@
...
+        -- e.g. if the mapping is made by PPSN is full, 
+        -- if thanks to name and birthday, partial
+        correlation_quality text NOT NULL DEFAULT 'full'
+    );
+
+CREATE TABLE
+    organisations_users (
+        organisation_id uuid NOT NULL,
+        user_id uuid NOT NULL REFERENCES users (id),
+        invitation_status text NOT NULL DEFAULT 'pending',
```
Thurfjell, line 18: Does this relate to the user_status in user table? 

Is there an on boarding journey to be an org user (is this the so called public servant?)?

packages/messages/migrations/sqls/20240520080749-add-users-data-up.sql
SamSalvatico: In future we can decide to make manual check for users if the quality is partial
If we matched it through PSSN we are sure that the organisation is importing the profile of a user that matches what we have in user profile
If only by name and first name we probably have to check before send an invitiation for the messaging platform to the email address

packages/messages/migrations/sqls/20240520080749-add-users-data-up.sql
SamSalvatico: yes, we send an invitation to join the messaging platform, this is related to that

packages/messages/migrations/sqls/20240520080749-add-users-data-up.sql
SamSalvatico: partially related
A user can accept to join the messaging platform but to decline to receive communications from an organisation

Thurfjell: What role does this table have?

Thurfjell: So is this user preferences, not users belonging to an organisation? 

I think it'd appreciate if we somehow made this clear, or am I missing some context?

SamSalvatico: it's both, it means that you have a relation with that org and you can select the transport for it

Thurfjell: I am talking about two different contexts. Are we mixing them?

1) User preferences. A user might have different preferences for different orgs.
2) Organisational user, a user who belongs to an organisation. This would probably be what we currently use "public servant" for. 

These two, imo, are highly separated concerns.

SamSalvatico: now I got it
Nope, this table is only related to the fact that a user can be contacted by an organisation

The fact that a user is "part of" an organisation will be managed in the user profile service

Thurfjell (approved):

PR #289, opened 2024-05-20
Feature: import users through CSV

SamSalvatico:
### Description
Added the API to make users able to import user-organizations relations through CSV file

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**


apps/messaging-api/services/users/csv-import.ts
```diff
@@ -0,0 +1,116 @@
...
+            users_data,
+            import_channel
+         values ($1, $2, $3)
+    `,
+      [organisationId, JSON.stringify(usersToImport), "csv"],
+    );
+  } catch (error) {
+    const toOutput = createError(
+      "SERVER_ERROR",
+      `Error during CSV file store on db: ${(error as Error).message}`,
```
Thurfjell, line 101: This should never fail right. I would rather use a guard (there's one util in scheduler proj) with like
```javascript
iserror(err: unknown) err is Error {
 return err instanceof Error && "message" in err
}
```
If not, write some string or something. Maybe just unnecessary fluff. 

You decide

apps/messaging-api/services/users/csv-import.ts
```diff
@@ -0,0 +1,116 @@
...
+  if (!file) {
+    throw new Error("file is missing");
+  }
+
+  const savedFiles = await req.saveRequestFiles();
+
+  return getUsersFromCsv(savedFiles[0].filepath);
+};
+
+export const importCsvFromRequest = async (params: {
```
Thurfjell, line 73: What's the value of having this as a separate function, rather than inline where it's used?

apps/messaging-api/services/users/csv-import.ts
```diff
@@ -0,0 +1,116 @@
...
+      500,
+    )();
+    params.req.log.error({ error: toOutput });
+    throw toOutput;
+  } finally {
+    client.release;
+  }
+};
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
Thurfjell, line 111: Isn't ts-expect-error to prefer?

apps/messaging-api/services/users/csv-import.ts
SamSalvatico: in this way we will be able to unit test it, in future

Thurfjell (approved):

PR #301, opened 2024-05-23
Chore: export different names for schemas and types

SamSalvatico:
### Description

To avoid conflicts I exported consts with different names for schemas and types

E.g. We used to export two const with name `AddressesList`, now we export `AddressesListSchema` and `AddressesList`
From
```
export const AddressesList = Type.Array(Address);
export type AddressesList = Static<typeof AddressesList>;
```
To
```
export const AddressesListSchema = Type.Array(AddressSchema);
export type AddressesList = Static<typeof AddressesListSchema>;
```

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [ ] **New feature**
- [x] **Dev change**

### Checklist:


edge33 (approved):
Great work, this should be a standard and we should adopt for every app

peschina (approved):
LGTM

PR #310, opened 2024-05-24
Feature: invoke user profile apis to map messaging user

SamSalvatico:
### Description

- Moved `building-blocks-sdk` to `es-module`
- Invoked `findUser` API in `messaging-api` to search for the imported users

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [x] **Dev change**

### Checklist:


SamSalvatico:
Hi all!

I had to set `"type": "module"` in `building-blocks-sdk` to make us able to import it in our APIs services
I just want to notify it to you, shouldn't disrupt Next services, I tested it locally also building images and it seems to work as expected

packages/building-blocks-sdk/package.json
```diff
@@ -8,5 +8,7 @@
   "devDependencies": {
     "@types/node": "20.10.4",
     "typescript": "5.3.3"
-  }
+  },
+  "type": "module",
```
edge33, line 12: good change, but we should also build this to plain js in the future.

edge33 (approved):
LGTM

packages/building-blocks-sdk/package.json
SamSalvatico: totally agree

Thurfjell (approved):
Lots of abstractions here. 

I'm always going to advocate to write a bit more boring code, without abstractions if possibly. I don't believe we need a JavaScript utility "findBy" when you can write each individual SQL query quite easily eg. It's always easier to come back and modify when it's just pure simpleton code, repeated or not.

PR #315, opened 2024-05-24
[Payments] Payments landing page update

peschina:
### Ticket:

- [16989](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/16989

### Description

Update landing page according to ticket

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**

### Checklist:

### Screenshots:

N/A

### DIFF
--- a/apps/payments/app/[locale]/info/Header.tsx
+++ b/apps/payments/app/[locale]/info/Header.tsx
@@ -27,7 +27,7 @@ export default async () => {
             }}
           >
             <a
-              href="/"
+              href="https://dev.blocks.gov.ie"
               className="govie-header__link govie-header__link--homepage"
               style={{ display: "block" }}
             >
--- a/apps/payments/app/[locale]/info/page.tsx
+++ b/apps/payments/app/[locale]/info/page.tsx
@@ -39,8 +39,16 @@ export default async (props: Props) => {
                 {t("AlphaBanner.tag")}
               </strong>
               <span className="govie-phase-banner__text">
-                {/* CHANGE HERE */}
-                {t("AlphaBanner.bannerText")}
+                {t.rich("AlphaBanner.bannerText", {
+                  anchor: (chunks) => (
+                    <a
+                      className="govie-link"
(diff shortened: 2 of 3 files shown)


apps/payments/app/[locale]/info/Header.tsx
```diff
@@ -27,7 +27,7 @@ export default async () => {
             }}
           >
             <a
-              href="/"
+              href="https://dev.blocks.gov.ie"
```
Comment (nnorbert) [inline apps/payments/app/[locale]/info/Header.tsx:30]: We should use the domain from an env variable so we would have control over this link on different environments.

apps/payments/app/[locale]/info/page.tsx
```diff
@@ -39,7 +39,16 @@ export default async (props: Props) => {
                 {t("AlphaBanner.tag")}
               </strong>
               <span className="govie-phase-banner__text">
-                {t("AlphaBanner.bannerText")}
+                {t.rich("AlphaBanner.bannerText", {
+                  anchor: (chunks) => (
+                    <a
+                      className="govie-link"
+                      href="https://www.formsg.testing.gov.ie/en/664c61ba5f7c9800231db294"
```
Comment (nnorbert) [inline apps/payments/app/[locale]/info/page.tsx:46]: Is this url okay to be hardcoded? Won't it be different per environment?

apps/payments/app/[locale]/info/Header.tsx
Comment (peschina) [reply apps/payments/app/[locale]/info/Header.tsx:30]: Good catch

apps/payments/app/[locale]/info/page.tsx
Comment (peschina) [reply apps/payments/app/[locale]/info/page.tsx:46]: I think it'll always be the same, but let me check

apps/payments/app/[locale]/info/Header.tsx
Comment (peschina) [reply apps/payments/app/[locale]/info/Header.tsx:30]: But we need to remember to have those vars updated when the building blocks site will be deployed, we currently have it in dev only

PR #326, opened 2024-05-27
Bootstrap forms landing page

alfonsograziano:
### Description

Just creating the structure for the new formsie landing page. 
The content will follow in a different PR.

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [X] **New feature**
- [ ] **Dev change**


apps/forms/app/[locale]/page.tsx
```diff
@@ -0,0 +1,18 @@
...
+import Header from "./Header";
+import Footer from "./Footer";
+import { getTranslations } from "next-intl/server";
+
+export default async function () {
+  const t = await getTranslations("LandingPage");
+
+  return (
+    <>
+      <Header />
```
Comment (nnorbert) [inline apps/forms/app/[locale]/page.tsx:13]: Shouldn't the Header/Footer be in the layout?

Comment (alfonsograziano) [reply apps/forms/app/[locale]/page.tsx:13]: We can keep it in the layout, but in this specific case, since it is just a single page there is no difference :) In case you prefer we can move it to the layout, not a big deal

Comment (nnorbert) [reply apps/forms/app/[locale]/page.tsx:13]: It's okay, if this will have only one page

nnorbert (approved):

PR #330, opened 2024-05-27
Feature(messaging): translatable messages

SamSalvatico:
### Description

Added a method to send messages based on the preferred language for the user

They will be used to send invitations

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**

### DIFF
--- a/apps/messaging-api/services/messages/messages.ts
+++ b/apps/messaging-api/services/messages/messages.ts
@@ -1,6 +1,8 @@
 import { PostgresDb } from "@fastify/postgres";
 import {
+  MessageInput,
   CreateMessage,
+  CreateTranslatableMessage,
   ReadMessage,
   ReadMessages,
 } from "../../types/schemaDefinitions";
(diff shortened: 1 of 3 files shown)


apps/messaging-api/services/messages/messages.ts
```diff
@@ -11,13 +14,62 @@ import { JobType } from "aws-sdk/clients/importexport";
...
 import { awsSnsSmsService } from "../sms/aws";
+//import { Profile } from "building-blocks-sdk";
 
 const EXECUTE_JOB_ERROR = "EXECUTE_JOB_ERROR";
 
+export const createTranslatableMessage = async (params: {
+  payload: CreateTranslatableMessage;
+  pg: PostgresDb;
+  requestUserId: string;
+}): Promise<void> => {
```
Thurfjell, line 25: Not so sure we have the concept of languages on non saved messages though. We'd need to treat it similarly to "templates" right. One message have many versions.

SamSalvatico: that's what happens thanks to the `CreateTranslatableMessage`

SamSalvatico:
Replaced by #334

PR #334, opened 2024-05-28
Feature(messaging): send invitations

SamSalvatico:
### Description

Send invitations to make the users join the messaging platform

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**


apps/messaging-api/services/users/import/send-invitations.ts
```diff
@@ -0,0 +1,302 @@
...
+      pg: params.pg,
+      organisationId: toImportUsers.organisationId,
+    });
+
+    await setImportedAsInvited({
+      invited: sent,
+      toImportUsers: params.toImportUsers,
+      pool: params.pg.pool,
+    });
+  } finally {
```
Thurfjell, line 61: Why do you need a transaction? Is the commit/rollback hidden somewhere?

SamSalvatico: it's not a transaction, using `client` prevents opening/closing connection multiple times if you know that you need to do queries sequentially

apps/messaging-api/services/users/import/send-invitations.ts
```diff
@@ -0,0 +1,302 @@
...
+    .map((ud) => ud.relatedUserProfileId) as string[];
+
+  if (importedUserProfileIds.length === 0) {
+    return;
+  }
+  const languagePerUser = getLanguagePerUser({
+    userIdsToSearchFor: importedUserProfileIds,
+    requestUserId: params.requestUserId,
+  });
+  const client = await pg.connect();
```
Thurfjell, line 37: Shouldn't this connection be grabbed from the pool?

apps/messaging-api/services/users/import/send-invitations.ts
```diff
@@ -0,0 +1,302 @@
...
+    });
+
+    const toSend = prepareInvitations({
+      userInvitations,
+      perIdLanguage: languagePerUser,
+    });
+
+    const sent = await sendInvitations({
+      toSend,
+      pg: params.pg,
```
Thurfjell, line 52: Why's the client not sent here? Do we need to hold it open until everything is finished? Does it depend on each other? If so use a transaction.

apps/messaging-api/services/users/import/send-invitations.ts
```diff
@@ -0,0 +1,302 @@
...
+      organisationId: toImportUsers.organisationId,
+    });
+
+    await setImportedAsInvited({
+      invited: sent,
+      toImportUsers: params.toImportUsers,
+      pool: params.pg.pool,
```
Thurfjell, line 59: same here

apps/messaging-api/services/users/import/send-invitations.ts
SamSalvatico: I had to do this because in the `createMessage` code we are directly using `pg` and it would take a while to refactor

PR #346, opened 2024-05-29
Move Building Blocks selector into a Shared components package

alfonsograziano:


apps/messaging/app/[locale]/Header.tsx
```diff
@@ -143,6 +145,8 @@ export default () => {
             >
               <ds.Icon icon="logout" color={ds.colours.ogcio.white} size={22} />
             </Link>
+
+            <BuildingBlockSelector locale={locale as "en" | "ga"} />
```
Comment (Thurfjell) [inline apps/messaging/app/[locale]/Header.tsx:149]: Would love if we didn't use the "as" type assertions

packages/shared-components/components/BuildingBlocksSelector/Selector.tsx
```diff
@@ -1,44 +1,60 @@
...
+    payments: "Payments",
+    messaging: "Messaging",
+    lifeEvents: "Life Events",
+  },
+  ga: {
+    payments: "Payments",
+    messaging: "Messaging",
+    lifeEvents: "Life Events",
+  },
+};
```
Thurfjell, line 43: 👌

Thurfjell (approved):

apps/messaging/app/[locale]/Header.tsx
Comment (alfonsograziano) [reply apps/messaging/app/[locale]/Header.tsx:149]: Mmm I agree, let me see if I can improve the type coming from the props :)

PR #348, opened 2024-05-29
17270 messaging landing page

nnorbert:
Messaging landing page.
Some other minor fixes.


apps/messaging/app/[locale]/admin/layout.tsx
```diff
@@ -0,0 +1,55 @@
+import { getTranslations } from "next-intl/server";
```
Comment (alfonsograziano) [inline apps/messaging/app/[locale]/admin/layout.tsx:1]: @nnorbert are we sure we need this change? Can you please provide a screenshot of before/after?

Thurfjell (approved):
🥇

apps/messaging/app/[locale]/layout.tsx
```diff
@@ -1,9 +1,5 @@
 import "design-system/dist/style.css";
```
Comment (alfonsograziano) [inline apps/messaging/app/[locale]/layout.tsx:1]: @Thurfjell since these changes will impact the Messaging application, do you mind taking a look? 
The Header and Footer components are copied from the other pages

alfonsograziano (approved):
LGTM overall

PR #443, opened 2024-05-30
Fix language switcher

nnorbert:
Fix language switcher

### DIFF
--- a/apps/payments/app/components/Header/LanguageSwitch.tsx
+++ b/apps/payments/app/components/Header/LanguageSwitch.tsx
@@ -1,6 +1,7 @@
 import ds from "design-system";
 import { headers } from "next/headers";
 import styles from "./Header.module.scss";
+import { redirect, RedirectType } from "next/navigation";
 
 type Theme = "dark" | "light";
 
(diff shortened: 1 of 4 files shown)


apps/payments/app/components/Header/LanguageSwitch.tsx
```diff
@@ -9,38 +10,72 @@ const colors: Record<Theme, string> = {
...
-          pathSlice.at(1) === "ga" ? "govie-!-font-weight-bold" : ""
-        }`.trim()}
-        style={{ color: colors[theme] }}
-        href={new URL("/ga/" + path, process.env.HOST_URL).href}
-      >
-        Gaeilge
-      </a>
-    </div>
+    <form>
+      <div className={styles.languagesContainer}>
```
hamza14khan, line 48: suggestion(non-blocking): an array that keeps state of selectors and then render it using loop.

e.g.

```
const languageOptions = [
{
 name: 'English',
 fontWeight: 700,
 handler: englishHandler,
 styles: ...
},
// same for Gaelge.
]

alfonsograziano (approved):
LGTM

apps/payments/app/components/Header/LanguageSwitch.tsx
nnorbert: Good idea, however, I don't think that this array should contain styling related things, because those are shared. I will push the changes soon.

nnorbert: Done

hamza14khan: Savage Norbert! 💪

hamza14khan (approved):

PR #512, opened 2024-05-31
Chore(messaging): invitation fixes

SamSalvatico:
### Description

Some fixex in invitation flow

## Type

- [ ] **Dependency upgrade**
- [x] **Bug fix**
- [ ] **New feature**
- [ ] **Dev change**

### DIFF
--- a/apps/messaging-api/routes/users/index.ts
+++ b/apps/messaging-api/routes/users/index.ts
@@ -1,6 +1,7 @@
 import { FastifyInstance, FastifyReply, FastifyRequest } from "fastify";
 import { Type } from "@sinclair/typebox";
 import {
+  IMPORT_USERS_ERROR,
   getCsvExample,
   importCsvFileFromRequest,
   importCsvRecords,
@@ -23,6 +24,7 @@ import {
   updateInvitationStatus,
   updateOrganisationFeedback,
 } from "../../services/users/invitations/accept-invitations";
+import { createError } from "@fastify/error";
 
 const tags = ["Users"];
 
@@ -44,8 +46,10 @@ export default async function users(app: FastifyInstance) {
       // exclamation mark used here because we have
       // verifyUser preValidation
       await importCsvFileFromRequest({
-        req: { ...request, user: request.user! },
+        filepath: await saveRequestFile(request),
+        user: request.user!,
         pg: app.pg,
+        logger: request.log,
       });
     },
   );
@@ -195,4 +199,19 @@ export default async function users(app: FastifyInstance) {
       }),
     }),
   );
+
+  const saveRequestFile = async (request: FastifyRequest): Promise<string> => {
+    const file = await request.files();
+    if (!file) {
+      throw createError(
+        IMPORT_USERS_ERROR,
+        "File is missing in the request",
+        400,
+      )();
+    }
+
+    const savedFiles = await request.saveRequestFiles();
+
+    return savedFiles[0].filepath;
+  };
 }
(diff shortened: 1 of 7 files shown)


apps/messaging-api/routes/users/index.ts
```diff
@@ -41,7 +41,12 @@ export default async function users(app: FastifyInstance) {
       },
     },
     async (request: FastifyRequest, _reply: FastifyReply) => {
-      await importCsvFileFromRequest({ req: request, pg: app.pg });
+      // exclamation mark used here because we have
+      // verifyUser preValidation
+      await importCsvFileFromRequest({
```
Thurfjell, line 46: If I were you, I would compose the relevant parameters to translate the csv format into rows in sql, instead of passing down something so abstract as a fustily request. 

So imagine like
```javascript
const something = await Something(req.whateverfile)
const csvData = await needMoreStuff(something)
await storeCsv(csvDat)
```

apps/messaging-api/routes/users/index.ts
```diff
@@ -115,7 +120,7 @@ export default async function users(app: FastifyInstance) {
       _reply: FastifyReply,
     ) => ({
       data: await getInvitationForUser({
-        userId: request.user!.id,
+        userProfileId: request.user!.id,
```
Thurfjell, line 123: I don't think user profile has an id column? It uses `user_id` iirc.

apps/messaging-api/services/tags/manage-tags.ts
```diff
@@ -46,7 +46,7 @@ const linkTagsToUser = async (params: {
         INSERT INTO tags_users (user_id, tag_id)
         VALUES ${toInsertIndexes.join(", ")}
         ON CONFLICT (user_id, tag_id)
-        DO NOTHING
+        DO UPDATE SET "user_id" = tags_users.user_id
```
Thurfjell, line 49: Why? 
Doesn't this overwrite all tag users?

apps/messaging-api/routes/users/index.ts
SamSalvatico: it has, but the authenticated user is related to the "centralized" user profile, then I have to use `user_profile_id` to match

apps/messaging-api/services/tags/manage-tags.ts
SamSalvatico: yes, with the same value, it is a workaround to make it returns all the related tags without having to run 2 queries

apps/messaging-api/routes/users/index.ts
Thurfjell: We're not talking about the profile.user_profiles now are we? 😬 Im so excited we have multiple user profile tables...

apps/messaging-api/services/tags/manage-tags.ts
Thurfjell: Just run two queries lol. Make a CTE. 😂

Thurfjell: This is not a high traffic area for sure. We can afford to play nice and dumb.

apps/messaging-api/routes/users/index.ts
SamSalvatico: Please, read the related code to understand what it does

apps/messaging-api/routes/users/index.ts
Thurfjell: But there's not really any service here, more than a collection of functions in a file. These service functions hard depend on request from a web framework. This is pretty tight coupling to begin with, no?

Thurfjell (approved):
You can merge as you wish. 👍

PR #519, opened 2024-06-03
feat(AuthService) redirect to subpath after login

edge33:
### Description

Currently, when users access the app via a subpath such as /timeline without being logged in, they are redirected to the authentication service. After authentication, they are then redirected back to the root level (/) of the app.

With this update, the authentication service will now redirect users back to the specific subpath they originally attempted to access. For example, if a user initially navigates to /timeline, after successful authentication, they will be redirected back to /timeline instead of the root level (/).

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**

### DIFF
--- a/packages/auth/src/route.ts
+++ b/packages/auth/src/route.ts
@@ -54,8 +54,8 @@ export default async (req: NextRequest) => {
   if (publicServant) {
     return redirect("/admin", RedirectType.replace);
   }
-  const redirectPath_ = req.nextUrl.searchParams.get("redirectPath");
-  const redirectUrl = redirectPath_ ? `${redirectPath_}` : "/";
+  const redirectPath = req.nextUrl.searchParams.get("redirectPath");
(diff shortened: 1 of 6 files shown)


packages/auth/src/route.ts
```diff
@@ -53,6 +54,8 @@ export default async (req: Request) => {
   if (publicServant) {
     return redirect("/admin", RedirectType.replace);
   }
+  const redirectPath_ = req.nextUrl.searchParams.get("redirectPath");
```
alfonsograziano, line 57: non blocking: is the `_` needed in the var name for some specific reason?

alfonsograziano (approved):
Super useful change! LGTM

packages/auth/src/route.ts
edge33: good catch!

PR #532, opened 2024-06-05
Feature(messaging): store user details

SamSalvatico:
### Description

Store the phone number, email, details to the user to manage the ones who do not exist on the user profile

Also fixed some tags queries
## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**

### DIFF
--- a/apps/messaging-api/services/tags/manage-tags.ts
+++ b/apps/messaging-api/services/tags/manage-tags.ts
@@ -3,7 +3,7 @@ import { Tag } from "../../types/usersSchemaDefinitions";
 import { createError } from "@fastify/error";
 
 const TAGS_PATH_SEPARATOR = ".";
-
+const MANAGE_TAGS_ERROR = "MANAGE_TAGS_ERROR";
 export const processTagsPerUser = async (params: {
   userId: string;
   tags: string[];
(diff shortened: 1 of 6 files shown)


apps/messaging-api/services/tags/manage-tags.ts
```diff
@@ -35,32 +34,34 @@ const linkTagsToUser = async (params: {
...
-    tagIds.push(toCreate.id);
-    tagIdsIndexes.push(`$${tagIdsIndex++}`);
-    toInsertIndexes.push(`($${indexCount++}, $${indexCount++})`);
+    toInsertValues.push(toCreate.id);
+    const currentTagIdIndex = indexCount++;
+    toInsertIndexes.push(`($1, $${currentTagIdIndex})`);
+    tagIdsIndexes.push(`$${currentTagIdIndex}`);
   }
 
+  await params.client.query(
```
SamSalvatico, line 46: Had to split query in two because using CTE does not return all the values from the `tags_users` table

apps/messaging-api/services/users/import/map-users.ts
```diff
@@ -56,12 +57,13 @@ const mapUsersSync = async (params: {
   const profile = new Profile(params.requestUser.id);
 
   const processingUsers = usersImport.usersData.map(
-    (toImportUser: ToImportUser) =>
-      processToImportUser({
+    async (toImportUser: ToImportUser) =>
+      await processToImportUser({
```
SamSalvatico, line 61: had to add await to avoid collisions across shared tags

Thurfjell (approved):

PR #555, opened 2024-06-07
[Payments] MyGovId stub for Logto

peschina:
### Ticket:

- [17562](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/17562)

### Description

Added mock endpoints for logto in `mock-api` and `auth-service` services.
The mocked flow is the following:
- login flow is initialised by logto
- user logs in choosing MyGovId account
- user is redirected to `http://localhost:8000/static/login/api/authorize-mock` and presented with a mock login page
- when the login form is submitted `http://localhost:8000/static/login/api/auth-mock` is called. The api creates a mock, signed jwt and redirects the user to logto callback, passing `state` and the id token as `code`
- logto calls our mock token endpoint on `http://mock-auth-service:3005/mock/token`, that will return the user tokens and info.
- logto calls our mock jwk set endpoint on `http://mock-auth-service:3005/mock/keys` to get the public key set and verify the signature of the token it just got from us
- logto verifies the identity of the user and signs in the user. If the user is not registered yet it is signed up first.

To use it locally:
- Create an RSA key pair on your machine with your preferred method. Add the private key as `JWK_PRIVATE_KEY` environment variable in `mock-api`. Add the public key as `JWK_PUBLIC_KEY` environment variable in `auth-service`. Be careful with newlines and their formatting :)
- Run `npm run start:logto`. This will spin up a dedicated docker compose with our services dependencies + logto, its database and the auth service. 
Please beware that at this stage the logto repo branch is no longer checkout out or pulled automatically, it is your responsibility to do so. The auth service is run in the docker compose and not locally with npm because it needs to be accessible by the logto application in docker. Port 3005 is still exposed though.
- Update MyGovId connector configuration (this should be reflected in the seed as well)
    - Authorization Endpoint: `http://localhost:3005/mock/auth`
    - Token Endpoint: `http://mock-auth-service:3005/mock/token`
    - Client id: same value as payments api application id
    - Client secret: same value as payments api application secret
    - Id Token verification config: 
```
{
     "jwksUri": "http://mock-auth-service:3005/mock/keys"
}
```
- Make sure that in Sign-in experience =>  Sign-up and sign-in =>`Enable user registration` is set to true

Issues to be addressed at some point:
- Redirection to the previous route is missing
- The implementation currently returns a mocked jwt with the payments client id audience. Depending on where the user wants to log in we should return the audience with the corresponding application id
- To allow registration of new users in logto the jwt returns a random `sub` and a random `mobile` (used for user identification with id and phone number). Atm to log in with a user that is already created you need to link the account when prompted (logto identity verification will find an account with same email, but different id and phone number). We might want to find a better solution to handle this
- env vars are spread across different services (for example the key pair to sign and verify the jwt)
- `mock-api` and `auth-service` services now provide two mocked flows to simulate a login with MyGovId. What is marked as "mock" is the mock used with logto, what is not marked as mock and was already there is actually the default mock. It is quite confusing, we should get back to this.

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**

### Checklist:

### Screenshots:

N/A


SamSalvatico (commented):
Thank you, @peschina !

Could you add some of the directives you wrote in the PR description in the README file? (Mainly the ones about the how-to setup)
And please also add some info on how to get the "logto app id" value

PR #566, opened 2024-06-10
Feature(messaging): settings APIs

SamSalvatico:
Description

Update APIs and SDKs for users settings

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**


apps/messaging-api/services/users/invitations/accept-invitations.ts
```diff
@@ -28,6 +31,35 @@ export const getInvitationForUser = async (params: {
...
+      userProfileId,
+      errorCode: ACCEPT_INVITATIONS_ERROR,
+    });
+    if (!user.userProfileId) {
+      throw createError(
+        ACCEPT_INVITATIONS_ERROR,
+        "The current user has no related user profile",
+        400,
+      )();
+    }
```
Thurfjell, line 52: Why is this needed? 
getUserInvitations seems to require the profile id. Should it not be able to determine based on this value and the credentials, what you should get in response?

apps/messaging-api/services/users/invitations/accept-invitations.ts
```diff
@@ -158,6 +190,42 @@ export const updateInvitationStatus = async (params: {
   }
 };
 
+export const getInvitationStatus = async (params: {
+  pg: PostgresDb;
+  userProfileId: string;
+}): Promise<{ userStatus: string }> => {
+  const client = await params.pg.connect();
```
Thurfjell, line 197: `pg.pool.connect()` ?

apps/messaging-api/services/users/invitations/accept-invitations.ts
```diff
@@ -158,6 +190,42 @@ export const updateInvitationStatus = async (params: {
...
+    throw toOutput;
+  } finally {
+    client.release();
+  }
+
+  if (statusResponse && statusResponse.rowCount) {
+    return statusResponse.rows[0];
+  }
+
+  throw createError(ACCEPT_INVITATIONS_ERROR, "Cannot find the user", 404)();
```
Thurfjell, line 226: I think you should revert this. 

We want the happy path at the end of the function, not the catastrophic one no? 

eg.

```
...
if(fail) throw
...
return happy
```

apps/messaging-api/services/users/invitations/accept-invitations.ts
SamSalvatico: You're totally right 🤦

Thurfjell (approved):

PR #569, opened 2024-06-11
Removed text from BB footers

Thurfjell:
### Ticket:

### Description

Removed all copy from building block footers (home, forms, messaging, payments)
Justified branding to end if larger than mobile, and centre if mobile. 

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [ ] **New feature**
- [ ] **Dev change**

### Checklist:

### Screenshots:

N/A


apps/home/app/[locale]/page.scss
```diff
@@ -38,6 +38,10 @@
   min-width: 0;
 }
 
+.custom-footer {
+  justify-content: end !important;
```
Comment (Thurfjell) [inline apps/home/app/[locale]/page.scss:42]: Not sure why important was needed here. The class precedence is in order for the element class. 🤷

SamSalvatico (approved):

PR #590, opened 2024-06-12
Logto integration with multiple applications

alfonsograziano:
This PR is a test and is not supposed to be delivered on PRODUCTION. 
We want to test how Logto behaves with multiple applications both on the login and on the logout side


apps/payments/app/[locale]/(hosted)/logto_integration/login/route.ts
```diff
@@ -0,0 +1,6 @@
+import { signIn } from "@logto/next/server-actions";
+import { logtoConfig } from "../page";
+
+export async function GET() {
+  await signIn(logtoConfig);
```
Comment (SamSalvatico) [inline apps/payments/app/[locale]/(hosted)/logto_integration/login/route.ts:5]: redirect uri is not requested here?

Comment (alfonsograziano) [reply apps/payments/app/[locale]/(hosted)/logto_integration/login/route.ts:5]: It is optional, the default is /callback

Comment (SamSalvatico) [reply apps/payments/app/[locale]/(hosted)/logto_integration/login/route.ts:5]: then if you login with payments or messaging logto redirects you to the correct app anyway?
Great

SamSalvatico (approved):

apps/payments/app/[locale]/(hosted)/logto_integration/login/route.ts
Comment (alfonsograziano) [reply apps/payments/app/[locale]/(hosted)/logto_integration/login/route.ts:5]: Yess, in Logto you still need to configure it with /callback, but if you don't require a specific redirect uri here, I think it picks the default one

PR #591, opened 2024-06-13
Fix messaging locale bugs in various urls

Thurfjell:
### Ticket:

### Description

Prefix most urls with locale prop. 
Changed some links to normal anchors so we make sure to have correct headers eg. and not lazy values from Link/next

## Type

- [ ] **Dependency upgrade**
- [x] **Bug fix**
- [ ] **New feature**
- [ ] **Dev change**

### Checklist:

### Screenshots:

N/A


apps/messaging/app/[locale]/admin/message-templates/TemplatesList.tsx
```diff
@@ -39,7 +39,7 @@ export default async () => {
                 className="govie-link govie-!-margin-right-3"
                 href={
                   urlWithSearchParams(
-                    templateRoutes.url,
+                    `${props.locale}/${templateRoutes.url}`,
                     { key: "id", value: template.templateMetaId },
                     { key: "lang", value: "en" },
```
Comment (SamSalvatico) [inline apps/messaging/app/[locale]/admin/message-templates/TemplatesList.tsx:44]: do we want to use `value: LANG_EN` here?

SamSalvatico (approved):

apps/messaging/app/[locale]/admin/message-templates/TemplatesList.tsx
Comment (Thurfjell) [reply apps/messaging/app/[locale]/admin/message-templates/TemplatesList.tsx:44]: Yup, let's change.

PR #593, opened 2024-06-13
[17562] MyGovID stub for Logto integration - refactored

peschina:
### Ticket:

- [17562](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/17562)

### Description

Added mock endpoints for logto in `mock-api` and `auth-service` services.
The mocked flow is the following:
- login flow is initialised by logto
- user logs in choosing MyGovId account
- user is redirected to `http://localhost:3005/logto/mock/auth` and presented with a mock login page
- when the login form is submitted `http://localhost:3005/logto/mock/login` is called. The api creates a mock, signed jwt and redirects the user to logto callback, passing `state` and the id token as `code`
- logto calls our mock token endpoint on `http://mock-auth-service:3005/mock/token`, that will return the user tokens and info.
- logto calls our mock jwk set endpoint on `http://mock-auth-service:3005/mock/keys` to get the public key set and verify the signature of the token it just got from us
- logto verifies the identity of the user and signs in the user. If the user is not registered yet it is signed up first.

To use it locally:
- Run `npm run start:logto`. This will spin up a dedicated docker compose with our services dependencies + logto, its database and the auth service. 
Please beware that at this stage the logto repo branch is no longer checkout out or pulled automatically, it is your responsibility to do so. The auth service is run in the docker compose and not locally with npm because it needs to be accessible by the logto application in docker. Port 3005 is still exposed though.
As an alternative you can run logto locally, making sure that its database is up and running.
Either way, be careful that the networking between logto and auth service is correctly set in the mock endpoints in the connector config (i.e. `localhost` vs `mock-auth-service`)
- Make sure that in Sign-in experience =>  Sign-up and sign-in =>`Enable user registration` is set to true

Issues to be addressed at some point:
- Redirection to the previous route is missing
- To allow registration of new users in logto the jwt returns a random `sub` and a random `mobile` (used for user identification with id and phone number). Atm to log in with a user that is already created you need to link the account when prompted (logto identity verification will find an account with same email, but different id and phone number). We might want to find a better solution to handle this

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**

### Checklist:

### Screenshots:

N/A


apps/auth-service/.env.sample
```diff
@@ -10,4 +10,4 @@ MYGOVID_URL=http://localhost:8000/static/login/api/authorize?redirect_uri=%CALLB
 CLIENT_SECRET=secret
 TOKEN_URL=http://localhost:8000/static/login/api/token?code=%CODE%
 ALLOWED_REDIRECT_HOSTS=^localhost(?::\d+)?$,\.*.gov.ie
-FALLBACK_REDIRECT_HOST=http://localhost:3000
\ No newline at end of file
+FALLBACK_REDIRECT_HOST=http://localhost:3000
```
nnorbert, line 13: New line is missing at the end.

alfonsograziano (approved):
LGTM - Thanks Sara!

nnorbert (approved):
LGTM!

edge33 (approved):
LGTM

PR #620, opened 2024-06-18
feat(Admin) Added search and page size to admin table

edge33:
### Description

- Added pagination and search users for admin panel
- Added prepared statements

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [x] **Dev change**

### Checklist:

### Screenshots:

N/A


apps/web/app/[locale]/admin/submissions/EventTable.tsx
```diff
@@ -47,26 +49,51 @@ export default async ({ searchParams, params }: SubmissionsTableProps) => {
...
+
+  userFlows = await postgres.pgpool.query(
+    `${dataSelect} ${dataQuery} ORDER BY updated_at DESC  LIMIT $${paramIndex++} OFFSET $${paramIndex}`,
+    [...sqlQueryParams, queryParams.limit, queryParams.offset],
   );
 
-  const countQueryResult = await postgres.pgpool.query<{
-    count: number;
-  }>(`SELECT COUNT(*) ${dataQuery}`);
+  countQueryResult = await postgres.pgpool.query(
```
Comment (alfonsograziano) [inline apps/web/app/[locale]/admin/submissions/EventTable.tsx:86]: non blocking: any reason to not declare this variable here directly instead of creating the let on top?

apps/web/app/[locale]/admin/submissions/components/TableControls/TableControls.tsx
```diff
@@ -22,12 +21,30 @@ export default async ({
...
+
+    const verifiedEmail = formData.get("verified-email") as string;
+    if (verifiedEmail.length > 0) {
+      searchParams.set("verifiedEmail", verifiedEmail);
+    }
+
     redirect(url.toString(), RedirectType.replace);
   };
 
+  const deviceType = filters.deviceType || "";
```
Comment (alfonsograziano) [inline apps/web/app/[locale]/admin/submissions/components/TableControls/TableControls.tsx:45]: non-blocking: Is filters.deviceType always there? Otherwise `{t("deviceType")}` might fail. In case it is always there, maybe is worth understanding why you need `|| ""`

alfonsograziano (approved):
I'm probably missing some business logic knowledge, but overall LGTM

SamSalvatico (approved):

apps/web/app/[locale]/admin/submissions/EventTable.tsx
Comment (edge33) [reply apps/web/app/[locale]/admin/submissions/EventTable.tsx:86]: good catch, this is indeed a leftover from a previous change

apps/web/app/[locale]/admin/submissions/components/TableControls/TableControls.tsx
Comment (edge33) [reply apps/web/app/[locale]/admin/submissions/components/TableControls/TableControls.tsx:45]: `{t("deviceType")}` uses the "deviceType" hardcoded string so it is not affected by deviceType retrieved from `filters`.

deviceType can be undefined if `?deviceType` queryParam is not present, it will preselect the "All" value for the device type filter select

Comment (alfonsograziano) [reply apps/web/app/[locale]/admin/submissions/components/TableControls/TableControls.tsx:45]: Uh my fault! Sorry :D

Comment (alfonsograziano) [reply apps/web/app/[locale]/admin/submissions/components/TableControls/TableControls.tsx:45]: I need to stop reviewing PRs before taking the first coffee in the morning :)

PR #621, opened 2024-06-18
Analytics integration

nnorbert:
Analytics integration


apps/home/.env.sample
```diff
@@ -1,3 +1,7 @@
 NEXT_TELEMETRY_DISABLED=1
 NODE_ENV=development
 ENVIRONMENT=DEV
+
+NEXT_PUBLIC_MATOMO_URL=
```
alfonsograziano, line 5: Shouldn't we add here an example of string here?

alfonsograziano, line 5: Since we need this variable: NEXT_PUBLIC_MATOMO_URL, do we need to change the Dockerfile and the other files to add the variables at build time? Otherwise the pipeline will fail

nnorbert: Done

nnorbert: Done. Terraform is also updated https://github.com/ogcio/life-events-terraform/pull/30

pipeline-variables/dev.yml
```diff
@@ -23,5 +23,6 @@ variables:
   logtoMessagingAppSectret: oTSELx7KrykDs8X57f4HOUni9eK91DNi
 
   nextPublicMatomoUrl: //analytics.ogcio.gov.ie
-  nextPublicMatomoSiteId: 11
   nextPublicMatomoProtocol: https
+  nextPublicLifeEventsMatomoSiteId: 11
+  nextPublicBuildingBlocksMatomoSiteId: 14
```
alfonsograziano, line 28: Do we want to have a single SiteId for all the Building Blocks? @andreacappadona17 this is more a question on the business side of things I think :)

apps/home/Dockerfile
```diff
@@ -5,15 +5,23 @@ ENV LOG_LEVEL=trace
...
+COPY ./packages/analytics/ /app/packages/analytics/
 
 FROM deps AS builder
 WORKDIR /app
 
+ARG NEXT_PUBLIC_MATOMO_URL
+ARG NEXT_PUBLIC_BUILDING_BLOCKS_MATOMO_SITE_ID
+ARG NEXT_PUBLIC_MATOMO_PROTOCOL
+ENV NEXT_PUBLIC_MATOMO_URL=$NEXT_PUBLIC_MATOMO_URL
+ENV NEXT_PUBLIC_MATOMO_SITE_ID=$NEXT_PUBLIC_BUILDING_BLOCKS_MATOMO_SITE_ID
```
alfonsograziano, line 23: I like this solution, makes sense :)

alfonsograziano (approved):
LGTM - @nnorbert shall we add some documentation as well to explain other devs how to integrate Matomo in their applications?

pipeline-variables/dev.yml
andreacappadona17: We'd rather have multiple siteIds, as it's possible to get aggregates of multiple sites by using roll-up reporting, while splitting metrics based on the url or source would be harder if we use the same website.
In terms of permissions, there could be a "payments analytics" permission that would be mapped to the single siteId.

andreacappadona17:
LGTM!

andreacappadona17 (approved):

documentation/docs/analytics.md
```diff
@@ -0,0 +1,102 @@
...
+
+### The AnalyticsTracker component
+
+This component must be placed in the page's layout at the top, right after the opening Body tag. It can take two optional parameters:
+- **userId**: this information can be sensitive; if used, it must comply with GDPR, and no sensitive data can be accessed based on this ID. This ID should not be a **PII** (Personally identifiable information), like `MyGovId`.
+- **customDimensions**: It is a custom attribute; we can have as many dimensions as we want. However, the current limit is 40 (dimension1 - dimension40). These attributes describe specific characteristics within your Matomo table report. Typically, dimensions are displayed in the first column of the report and provide context to the metrics.
```
andreacappadona17, line 34:
```suggestion
- **customDimensions**: It is a custom attribute; we can have as many dimensions as we want. However, the current limit is 40 (dimension1 - dimension40). These attributes describe specific characteristics within your Matomo table report. Typically, dimensions are displayed in the first column of the report and provide context to the metrics. Dimensions must be configured on Matomo by admins before using them in the code.
```

andreacappadona17 (changes requested):

andreacappadona17 (approved):

PR #639, opened 2024-06-24
one from address per organisation

Thurfjell:
### Ticket:

- [18087](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/18087)

### Description

Prevent users to upsert email providers with same from address per organisation.

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [ ] **New feature**
- [ ] **Dev change**

### Checklist:

### Screenshots:

N/A

### DIFF
--- a/apps/messaging-api/routes/providers/emails.ts
+++ b/apps/messaging-api/routes/providers/emails.ts
@@ -1,7 +1,15 @@
 import { Type } from "@sinclair/typebox";
 import { FastifyInstance } from "fastify";
 import { EmailProvider, mailService } from "./services";
-import { NotFoundError, ServerError } from "shared-errors";
+import {
+  isLifeEventsError,
+  NotFoundError,
+  ServerError,
+  ValidationError,
+} from "shared-errors";
+import { organisationId } from "../../utils";
+import { isNativeError } from "util/types";
+import { HttpError } from "../../types/httpErrors";
 const tags = ["Providers - Emails"];
 
 const EMAIL_PROVIDER_ERROR = "EMAIL_PROVIDER_ERROR";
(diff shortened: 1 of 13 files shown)


apps/messaging-api/routes/providers/emails.ts
```diff
@@ -58,16 +66,23 @@ export default async function emails(app: FastifyInstance) {
...
+      const client = await app.pg.pool.connect();
       const service = mailService(client);
       try {
-        return { data: await service.getProviders() };
+        return { data: await service.getProviders(organisationId) };
+      } catch (err) {
+        throw new ValidationError(EMAIL_PROVIDER_ERROR, "", [
+          {
+            fieldName: "provider",
+            message: "in_user",
```
SamSalvatico, line 83:
```suggestion
            message: "alreadyInUse",
```
As you wrote in other cases

apps/messaging-api/routes/providers/emails.ts
```diff
@@ -187,16 +196,17 @@ export default async function emails(app: FastifyInstance) {
...
     },
     async function handler(request, _reply) {
-      const client = await app.pg.connect();
-      const service = await mailService(client);
+      const client = await app.pg.pool.connect();
+      const service = mailService(client);
+
       try {
-        await service.deleteProvider(request.params.providerId);
+        await service.deleteProvider(organisationId, request.params.providerId);
```
SamSalvatico, line 209: In the `updateProvider` you are not catching, could be useful to use the same approach in both cases?

apps/messaging-api/routes/providers/services.ts
```diff
@@ -23,40 +24,115 @@ export interface SendMailParams {
...
+          ssl,
+          organisationId,
+        ],
+      );
+
+      const id = insertQueryResult.rows.at(0)?.id;
+
+      if (!id) {
+        throw new ServerError(
+          "EMAIL_PROVIDER_ERROR",
```
SamSalvatico, line 105: please set `EMAIL_PROVIDER_ERROR` in a const to reuse it

apps/messaging-api/routes/providers/emails.ts
Thurfjell: Thanks, let's use camel to make translations straight forward? Would be nice with some type/const package but let's not worry about that now.

Thurfjell: Actually forgot to remove this one 😇

apps/messaging-api/routes/providers/emails.ts
Thurfjell: There's no catching in create either. If there's any database fetch error, we bubble it to server error through the magic fastify error wrapper. 

We only specifically throw the variable error if you try to set forbidden values.

Thurfjell: Aha, no let's remove it.

SamSalvatico (approved):

PR #640, opened 2024-06-24
chore(admin) improved admin route protection and redirects

edge33:
### Description

- Improved route protection and admin redirects

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [ ] **New feature**
- [x] **Dev change**

### Checklist:

### Screenshots:

N/A


apps/web/app/[locale]/[event]/[...action]/page.tsx
```diff
@@ -39,4 +40,6 @@ export default async function ActionPage(props: web.NextPageProps) {
   }
 
   throw notFound();
-}
+};
+
+export default PublicFacingPage(Page);
```
Comment (alfonsograziano) [inline apps/web/app/[locale]/[event]/[...action]/page.tsx:45]: non-blocking: in case you want, you can achieve the same result by putting the PublicFacingPage logic in the layout file instead of wrapping the page component

alfonsograziano (approved):
LGTM

SamSalvatico (approved):

apps/web/app/[locale]/[event]/[...action]/page.tsx
Comment (edge33) [reply apps/web/app/[locale]/[event]/[...action]/page.tsx:45]: I tried that, in fact route protection under the admin path is achieved on the layout level.
Admin layout is a child of root layout, and for handling protection on the root layout I would have to access the requested url in the layout? do you think that is feasible?

apps/web/app/[locale]/[event]/[...action]/GetDigitalWallet/DeviceSelection.tsx
```diff
@@ -90,7 +90,78 @@ export default async (props: {
     );
 
     return redirect(urlBase);
-  }
+  });
+
+  // async function submitAction(async (formData: FormData) {
```
Comment (alfonsograziano) [inline apps/web/app/[locale]/[event]/[...action]/GetDigitalWallet/DeviceSelection.tsx:95]: Non-blocking: do you need to keep all this commented code?

apps/web/app/utils/redirectPublicServant.ts
```diff
@@ -0,0 +1,9 @@
+import { PgSessions } from "auth/sessions";
```
alfonsograziano, line 1: Non-blocking: I cannot find any usage of this one in the PR

edge33: right, it is not needed anymore

apps/web/app/[locale]/[event]/[...action]/GetDigitalWallet/DeviceSelection.tsx
Comment (edge33) [reply apps/web/app/[locale]/[event]/[...action]/GetDigitalWallet/DeviceSelection.tsx:95]: thanks, I forgot to remove

alfonsograziano (approved):

PR #641, opened 2024-06-24
chore(docker): add script with scheduler

SamSalvatico:
### Description

To speed up common startup of the service, removed the scheduler from the normal start script
## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [ ] **New feature**
- [x] **Dev change**


Thurfjell:
I mean anything relying on scheduling wont work if you `run npm start ` now. Is there no way of caching/speeding up the image? (always is isn't it). This will break the "one script to run app".

Thurfjell (changes requested):
Let's discuss implications.

SamSalvatico:
> I mean anything relying on scheduling wont work if you `run npm start ` now. Is there no way of caching/speeding up the image? (always is isn't it). This will break the "one script to run app".

Probably you're right, I will make the default script also run scheduler and create one with no scheduler

Thurfjell (approved):

PR #645, opened 2024-06-25
Feature(nextjs-logging-wrapper): added package

SamSalvatico:
### Description

Added the `nextjs-logging-wrapper` package

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**


Thurfjell (approved):

packages/nextjs-logging-wrapper/src/server-logger.ts
```diff
@@ -0,0 +1,97 @@
...
+    logger = getPinoInstance({
+      getConfigurations: getLoggerConfiguration,
+      minLevel: minimumLevel,
+    });
+  }
+
+  return logger;
+};
+
+export const getLoggerConfiguration = (
```
edge33, line 35: `getLoggerConfiguration` looks the same on both client and server logger, if this is the case, would it make sense to externalise this config to a separate file and import it?

SamSalvatico: they are slightly different, but I can try to normalize them 🆗

edge33: if they are different it is not worth externalising then. thanks for sharing, LGTM

edge33 (approved):
LGTM

PR #663, opened 2024-07-03
Feature(api-auth): Parse context to make it easily usable

SamSalvatico:
### Ticket:

- [18959](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/18959)

### Description

Now we parse the `LogtoContext` to make easier to work with that, instead of having different format if citizen/public servant, now they share a common interface

**Please Note**: we must maintain the two different request to get the context as a citizen or as a public servant because we need to differentiate requested scopes

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**

### Checklist:

### Screenshots:

`Bob Red` image is an example of what is returned if the user is a citizen, other one is for public servant
Screenshot 2024-07-03 at 16 44 22
Screenshot 2024-07-03 at 16 53 46


edge33 (approved):
LGTM

peschina (approved):
LGTM

nnorbert (approved):
LGTM!

PR #664, opened 2024-07-03
feat(upload-api) Added new upload api service

edge33:
### Ticket:

- [18782](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_boards/board/t/Life%20Events/Stories?System.AreaPath=Life%20Events&workitem=18782)

### Description

Initial versione of file upload service

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**

### Checklist:

### Screenshots:

N/A


apps/file-upload-api/openapi-definition.yml
```diff
@@ -0,0 +1,54 @@
...
+  version: 0.1.0
+components:
+  schemas: {}
+paths:
+  /health:
+    get:
+      responses:
+        "200":
+          description: Default Response
+  /files/upload:
```
relvao, line 14: why not just /files?

apps/file-upload-api/src/app.ts
```diff
@@ -0,0 +1,88 @@
...
+
+export async function build(opts?: FastifyServerOptions) {
+  const app = fastify(opts).withTypeProvider<TypeBoxTypeProvider>();
+  initializeLoggingHooks(app);
+  initializeErrorHandler(app);
+
+  // limit 20 gb file
+  app.register(multipart, {
+    limits: {
+      fileSize: 20 * 1024 * 1024 * 1024,
```
relvao, line 30: can we add the file limit to an env var?

apps/file-upload-api/src/app.ts
```diff
@@ -0,0 +1,88 @@
...
+      ),
+    },
+  });
+
+  app.register(postgres, {
+    host: app.config.POSTGRES_HOST,
+    port: Number(app.config.POSTGRES_PORT),
+    user: app.config.POSTGRES_USER,
+    password: app.config.POSTGRES_PASSWORD,
+    database: app.config.POSTGRES_DB_NAME_SHARED,
```
relvao, line 70: we need to create a db for this project. Shouldn't use the shared db. Happy for it to be a follow up story

apps/file-upload-api/src/config.ts
```diff
@@ -0,0 +1,12 @@
...
+  type: "object",
+  required: ["CLAMAV_HOST", "S3_ENDPOINT", "S3_REGION", "S3_BUCKET_NAME"],
+  properties: {
+    CLAMAV_HOST: { type: "string" },
+    S3_ENDPOINT: { type: "string" },
+    S3_REGION: { type: "string" },
+    S3_ACCESS_KEY_ID: { type: "string" },
+    S3_SECRET_ACCESS_KEY: { type: "string" },
+    S3_BUCKET_NAME: { type: "string" },
+  },
```
relvao, line 11: shouldn't the db config be a part of this as well?

apps/file-upload-api/src/routes/files/index.ts
```diff
@@ -0,0 +1,120 @@
...
+      return reject(app.httpErrors.badRequest("Filename not provided"));
+    }
+
+    const s3Config = app.s3Client;
+    let scanComplete = false;
+    let outputFinished = false;
+    let fileInfected = false;
+    const antivirusPassthrough = app.avClient.passthrough();
+    antivirusPassthrough.on("error", (err) => {
+      reject(app.httpErrors.internalServerError(err.message));
```
Eomm, line 35: I think this error would be bubble up to the pipeline error too - this means that we would run `reject` twice.

In this case, you may handle it with an internal `EventEmitter` and the `once` -> so you centralize who is gonna to run `resolve/reject`

apps/file-upload-api/src/app.ts
Eomm, line 30: Did you set the https://fastify.dev/docs/latest/Reference/Server/#requesttimeout ?

In this case I think you need to customize it with `request.raw.setTimeout()` for the upload routes

apps/file-upload-api/src/routes/files/index.ts
edge33: I like this approach but still, there are cases where `antivirusPassthrough` will receive multiple `error` events, if we catch an error with `once` subsequent errors will be thrown. It is also the case that even if antivirus fails the pipeline will push the stream data to the upload that will fail in consequence. I might accept having multiple error logs, what do you think?

apps/file-upload-api/src/routes/files/index.ts
```diff
@@ -0,0 +1,366 @@
...
+        app.log.error(err);
+      });
+
+      reply.send(downloadPassthrough);
+
+      const thePromise = new Promise<void>((resolve, reject) => {
+        antivirusPassthrough.once("scan-complete", (result) => {
+          const { isInfected } = result;
+          if (isInfected) {
+            reject("File is infected");
```
Eomm, line 301:
```suggestion
            reject("File is infected");
            return;
```

apps/file-upload-api/src/routes/files/index.ts
```diff
@@ -0,0 +1,366 @@
...
+      const monitorPassThrough = new PromiseTransform(thePromise);
+
+      pipeline(
+        stream,
+        antivirusPassthrough,
+        monitorPassThrough,
+        downloadPassthrough,
+        (err) => {
+          if (err) {
+            downloadPassthrough.destroy();
```
Eomm, line 315: I would expect it is not necessary

https://nodejs.org/api/stream.html#streampipelinestreams-callback

stream.pipeline() will call stream.destroy(err) on all streams except:

Readable streams which have emitted 'end' or 'close'.
Writable streams which have emitted 'finish' or 'close'.

apps/file-upload-api/.gitignore
```diff
@@ -0,0 +1,3 @@
+node_modules
+.env
+.tap
```
SamSalvatico, line 3: could you have a `dist` folder?

apps/file-upload-api/src/app.ts
```diff
@@ -0,0 +1,96 @@
...
+    limits: {
+      fileSize: app.config.MAX_FILE_SIZE as number,
+    },
+  });
+
+  app.register(fastifySwagger, {
+    openapi: {
+      info: {
+        title: "OGCIO File Upload API",
+        description: "API for OGCIO file upload servce",
```
SamSalvatico, line 51:
```suggestion
        description: "API for OGCIO file upload service",
```

apps/file-upload-api/src/config.ts
```diff
@@ -0,0 +1,19 @@
...
+    "MAX_FILE_SIZE",
+  ],
+  properties: {
+    CLAMAV_HOST: { type: "string" },
+    S3_ENDPOINT: { type: "string" },
+    S3_REGION: { type: "string" },
+    S3_ACCESS_KEY_ID: { type: "string" },
+    S3_SECRET_ACCESS_KEY: { type: "string" },
+    S3_BUCKET_NAME: { type: "string" },
+    MAX_FILE_SIZE: { type: "number" },
```
SamSalvatico, line 17: you are missing the `LOGTO_` keys

apps/file-upload-api/src/index.ts
```diff
@@ -0,0 +1,17 @@
...
+import { getLoggingConfiguration } from "logging-wrapper";
+
+const app = await build(getLoggingConfiguration());
+
+app.listen({ host: "0.0.0.0", port: 8006 }, (err, address) => {
+  if (err) {
+    console.error(err);
+    process.exit(1);
+  }
+  console.log(`app listening at ${address}`);
```
SamSalvatico, line 13:
```suggestion
  app.log.info(`app listening at ${address}`);
```

apps/file-upload-api/src/routes/files/index.ts
```diff
@@ -0,0 +1,362 @@
...
+      Key: filename,
+    }),
+  );
+};
+
+const scanAndUplad = async (app: FastifyInstance, request: FastifyRequest) => {
+  const data = await request.file();
+
+  if (!data) {
+    throw app.httpErrors.badRequest("Request is not multipart");
```
SamSalvatico, line 46: please use the shared errors
`new BadRequest( ....`

apps/file-upload-api/src/routes/files/index.ts
```diff
@@ -0,0 +1,362 @@
...
+  }
+
+  const stream = data.file;
+  const filename = data.filename;
+
+  if (!filename) {
+    throw app.httpErrors.badRequest("Filename not provided");
```
SamSalvatico, line 53: same

apps/file-upload-api/src/routes/files/index.ts
```diff
@@ -43,35 +53,36 @@ const scanAndUplad = async (app: FastifyInstance, request: FastifyRequest) => {
...
+      reject(new CustomError(FILE_UPLOAD, "File is infected", 400));
     });
 
     eventEmitter.once("fileUploaded", () => {
       resolve();
     });
 
     eventEmitter.once("fileTooLarge", () => {
-      reject(app.httpErrors.badRequest("File too large"));
+      reject(new CustomError(FILE_UPLOAD, "File is too large", 400));
```
SamSalvatico, line 78: why you use Custom Error instead of BadRequest?

edge33: good catch!

SamSalvatico (approved):

PR #667, opened 2024-07-08
Feature(messaging): migrations from UUID to nanoid

SamSalvatico:
### Description

Created the migrations to move user id and organisation id from UUID to text in messaging and profile.

Why? Because we will store user id and organisation id from Logto, that uses the `nanoId` format

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [ ] **New feature**
- [x] **Dev change**


apps/profile-api/.env.sample
SamSalvatico, line 1: We are seeding `webhooks_local_signing_key` in the logto seeder

apps/profile/migrations/sqls/20240708133834-from-uuid-to-text-down.sql
```diff
@@ -0,0 +1 @@
+/* Replace with your SQL commands */
```
alfonsograziano, line 1: Is it fine to have an empty down migration? I don't expect us to revert this PR, just something to keep in mind

apps/profile/migrations/sqls/20240708133834-from-uuid-to-text-up.sql
```diff
@@ -0,0 +1,23 @@
+BEGIN;
+
+ALTER TABLE user_addresses
+DROP CONSTRAINT user_addresses_user_id_fkey;
+
+ALTER TABLE user_entitlements
+DROP CONSTRAINT entitlements_user_id_fkey;
+
+ALTER TABLE user_details
+ALTER COLUMN user_id SET DATA TYPE varchar(12) USING SUBSTRING(user_id::text, 0, 12);
```
alfonsograziano, line 10: Uh that's an interesting approach! In payments, we're just truncating the tables, but also creating a substring is a good idea. Also, the risk of collisions is super low based on the current load of the non-prod envs. @nnorbert maybe something to keep in mind for next time :)

alfonsograziano (approved):
LGTM

PR #668, opened 2024-07-08
feat: change userid type to text

nnorbert:
feat: change userid type to text


apps/payments/migrations/sqls/20240708133624-change-user-id-type-down.sql
```diff
@@ -0,0 +1,5 @@
+/* Replace with your SQL commands */
+
+ALTER TABLE payment_providers ALTER COLUMN user_id TYPE uuid;
+ALTER TABLE payment_requests ALTER COLUMN user_id TYPE uuid;
+ALTER TABLE payment_transactions ALTER COLUMN user_id TYPE uuid;
```
alfonsograziano, line 5: EOF missing in both sql files :)

alfonsograziano (approved):
LGTM - just need to add the EOF

PR #671, opened 2024-07-09
Feature(messaging): integrate with logto

SamSalvatico:
### Description

Integrate messaging with Logto

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**


apps/messaging/app/[locale]/UserIcon.tsx
```diff
@@ -1,9 +1,12 @@
-import { PgSessions } from "auth/sessions";
 import ds from "design-system";
+import { getAuthenticationContext } from "./logto_integration/config";
```
Comment (alfonsograziano) [inline apps/messaging/app/[locale]/UserIcon.tsx:2]: non-blocking: ideally we want to remove the `logto_integration` folder soon, maybe the config can be moved in another folder (see the payments PR for example)

apps/messaging/app/[locale]/admin/PageWithMenuFlexWrapper.tsx
```diff
@@ -19,9 +19,9 @@ export default async ({ children }: { children: React.ReactNode }) => {
     <div style={{ display: "flex", gap: "30px", width: "100%" }}>
       <SideMenu
         locale={headers().get("x-next-intl-locale")?.toString() || LANG_EN}
-        options={await messages.sideMenuOptions(publicServant)}
+        options={await messages.sideMenuOptions(isPublicServant)}
         selected={selected}
-        userName={`${firstName} ${lastName}`}
+        userName={user.name ?? ""}
```
Comment (alfonsograziano) [inline apps/messaging/app/[locale]/admin/PageWithMenuFlexWrapper.tsx:24]: non-blocking: at some point, shall we make `name` a mandatory string? I cannot see any case on top of my head where we don't have the name field populated. Something to look into another PR though

apps/messaging/app/[locale]/admin/message-events/page.tsx
```diff
@@ -70,8 +71,9 @@ export default async (props: { searchParams: { search?: string } }) => {
   }
 
   const freeSearch = props.searchParams.search;
-  const { userId } = await PgSessions.get();
-  const client = new Messaging(userId);
+  const { accessToken } = await getAuthenticationContext();
```
Comment (alfonsograziano) [inline apps/messaging/app/[locale]/admin/message-events/page.tsx:74]: non-blocking: @nnorbert used a factory function to generate an sdk instance with a valid access token in a single call (just to make the code more readable). You might consider doing this type of refactoring as well. It is not mandatory though, is perfectly fine to keep all the calls like this

apps/messaging/app/[locale]/admin/PageWithMenuFlexWrapper.tsx
Comment (SamSalvatico) [reply apps/messaging/app/[locale]/admin/PageWithMenuFlexWrapper.tsx:24]: I really would like to, but we have to ensure that Logto is sending it to us, we can do the same checks as we do for the accessToken

apps/messaging/app/[locale]/logto_integration/config.ts
```diff
@@ -39,12 +43,31 @@ export default {
...
-    const citizenContext = await getCitizenContext();
-    if (citizenContext.isPublicServant) {
-      return getPublicServantContext();
+    let context = await getCitizenContext();
+    if (context.isPublicServant) {
+      context = await getPublicServantContext();
     }
 
-    return citizenContext;
+    if (!context.accessToken) {
```
Comment (alfonsograziano) [inline apps/messaging/app/[locale]/logto_integration/config.ts:51]: @nnorbert most probably this is something we want to do as well :)

apps/messaging/app/[locale]/logto_integration/config.ts
```diff
@@ -39,12 +43,31 @@ export default {
 
 export const getAuthenticationContext =
   async (): Promise<AuthSessionContext> => {
-    const citizenContext = await getCitizenContext();
-    if (citizenContext.isPublicServant) {
-      return getPublicServantContext();
+    let context = await getCitizenContext();
+    if (context.isPublicServant) {
+      context = await getPublicServantContext();
```
Comment (alfonsograziano) [inline apps/messaging/app/[locale]/logto_integration/config.ts:48]: (same applies for payments) I'm just a little bit concerned by these two calls repeated every time. I'm not sure if the result is cached or something in the SDK. We should take a look

Comment (SamSalvatico) [reply apps/messaging/app/[locale]/logto_integration/config.ts:48]: I will reuse your factory approach for token and also for this

apps/messaging/app/[locale]/admin/message-events/page.tsx
Comment (SamSalvatico) [reply apps/messaging/app/[locale]/admin/message-events/page.tsx:74]: I like your approach, will reuse it

apps/messaging/app/utils/routes.ts
```diff
@@ -51,6 +51,14 @@ export const userOrganisationsRoutes = {
   url: "settings/organisations",
 };
 
+export const logtoLogin = {
+  url: "/logto_integration/login",
```
alfonsograziano, line 55: You can keep login/logout at this endpoint if you prefer. I'd personally change it to `/login` and `/signout` without the `logto_integration` prefix

apps/messaging/app/[locale]/logto_integration/config.ts
Comment (alfonsograziano) [reply apps/messaging/app/[locale]/logto_integration/config.ts:48]: The factory doesn't fix this issue unfortunately. As far as I can see, it looks like we're doing two calls as well. 
But again, the factory is just an option, what you did is 100% working and adding further complexity might not be worth the time invested :D

alfonsograziano (approved):
Apart from a couple of minor non-blocking comments, it LGTM!

PR #672, opened 2024-07-10
Feature(profile-api): save user profile from webhook

SamSalvatico:
### Description

Now, when the webhook is invoked, the users are upserted in the `user-profile` db

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**


apps/profile-api/services/users/create-user.ts
```diff
@@ -0,0 +1,58 @@
...
+  ]);
+
+  let index = 2;
+  const indexes: string[] = ["$1"];
+  const values: (string | number | boolean | null)[] = [userId];
+  const columns: string[] = ["user_id"];
+  const toMapObject = createUserData as {
+    [key: string]: string | number | boolean | null;
+  };
+  for (const column in inputToColumnsMapping) {
```
alfonsograziano, line 43: non-blocking: I got what you're doing here, but it took me a little bit to get it. Wouldn't be possible to just add everything inside the query itself instead of generating both the columns and indexes at runtime?

alfonsograziano (approved):
LGTM

PR #673, opened 2024-07-10
feat(logto) Life events logto admin

edge33:
### Description

Integrates logto for admin users in life events

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**

### Checklist:

### Screenshots:

N/A


apps/web/.env.sample
```diff
@@ -28,3 +28,9 @@ NEXT_PUBLIC_MATOMO_URL=
 NEXT_PUBLIC_MATOMO_SITE_ID=
 NEXT_PUBLIC_MATOMO_PROTOCOL=http
 ENVIRONMENT=DEV
+
+NEXT_PUBLIC_LIFE_EVENTS_ENTRY_POINT=http://localhost:3000
```
alfonsograziano, line 32: Have you included these variables in Docker and in the pipeline variables?

apps/web/app/[locale]/(public)/page.tsx
```diff
@@ -9,6 +9,8 @@ type Props = {
 };
 
 export default async (props: Props) => {
+  console.log("page");
```
Comment (alfonsograziano) [inline apps/web/app/[locale]/(public)/page.tsx:12]: Let's remove this CL :)

apps/web/app/[locale]/admin/logto_integration/config.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    // if (citizenContext.isPublicServant) {
+    return getPublicServantContext();
+    // }
+
+    // return citizenContext;
+  };
+
+// unused for now, will add back when adding logto for citizens
+const getCitizenContext = () =>
```
Comment (alfonsograziano) [inline apps/web/app/[locale]/admin/logto_integration/config.ts:49]: Shall we just delete this function for now? I think we might need to rethink the getAuthContext to avoid too many calls to Logto, so most probably we can safely remove it from here for now

apps/web/app/[locale]/admin/logto_integration/config.ts
```diff
@@ -0,0 +1,80 @@
+import { AuthSession, AuthUserScope } from "auth/auth-session";
+import { AuthSessionContext } from "auth/types";
+
+// export const messagingApiResource = process.env.MESSAGES_BACKEND_URL + "/";
```
Comment (alfonsograziano) [inline apps/web/app/[locale]/admin/logto_integration/config.ts:4]: You can remove this comment since we don't have any API Resource in Life events

apps/web/app/[locale]/admin/layout.tsx
```diff
@@ -1,12 +1,54 @@
...
+      <body
+        style={{
+          margin: "unset",
+          minHeight: "100vh",
+          position: "relative",
+          display: "flex",
+          flexDirection: "column",
+        }}
+      >
+        {/* <AnalyticsTracker
```
Comment (alfonsograziano) [inline apps/web/app/[locale]/admin/layout.tsx:35]: Any specific reason to keep this AnalyticsTracker here commented? Do we need it?

alfonsograziano:
Just a last comment on top of my head: since the vast majority of this PR involves moving files and updating imports to separate the two layouts, I'd personally create and merge a PR just to do that in preparation, so that the Logto integration PR only contains the minimum changes required

apps/web/app/[locale]/admin/logto_integration/config.ts
```diff
@@ -0,0 +1,80 @@
...
+const organizationId = "ogcio";
+
+const orgScopes = [
+  AuthUserScope.Organizations,
+  AuthUserScope.OrganizationRoles,
+];
+
+// TODO: TBD
+const citizenScopes = [];
+const publicServantScopes = ["life-events:digital-wallet-flow:*"];
```
Comment (alfonsograziano) [inline apps/web/app/[locale]/admin/logto_integration/config.ts:25]: You're requesting the scope, but then when I'm authenticated in the app, you're never validating that I have this scope. Basically, with the current setup, we're not enforcing that the user has the rights to perform the required actions

apps/web/.env.sample
edge33: right right, out of curiosity, who do we need it to be `NEXT_PUBLIC` ? are we using anywhere client side?

alfonsograziano: Not in this case, but for the other variables (like the one in messaging and payments) we were using the same variable both for logto and for the building blocks selector (which is exposed client side as well). In your case you can remove the NEXT_PUBLIC if you want I think

apps/web/app/[locale]/admin/layout.tsx
Comment (edge33) [reply apps/web/app/[locale]/admin/layout.tsx]: thanks, removed it

apps/web/app/[locale]/admin/logto_integration/config.ts
Comment (alfonsograziano) [reply apps/web/app/[locale]/admin/logto_integration/config.ts:23]: In api-auth we have a checkPermissions plugin. We can reuse the same logic to extract the scopes, validate the issuer and validate that the user has the right scopes to perform the actions

peschina (approved):
LGTM

apps/web/app/[locale]/admin/submissions/page.tsx
```diff
@@ -27,6 +31,18 @@ const componentsMap: {
 };
 
 export default async (props: SubmissionsTableProps) => {
+  const authFactory = AuthenticationFactory.getInstance();
+  const context = await authFactory.getPublicServant();
+
+  const hasPermissions = hasAdminPermissions(
```
Comment (SamSalvatico) [inline apps/web/app/[locale]/admin/submissions/page.tsx:37]: I think that this can be replaced by `authFactory.isPublicServant`, but as you prefer

SamSalvatico (dismissed):

SamSalvatico (approved):

PR #676, opened 2024-07-10
Chore/auth share factories

SamSalvatico:
### Description

Share the authentication context management

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [] **New feature**
- [x] **Dev change**


packages/nextjs-logging-wrapper/src/common-logger.ts
SamSalvatico, line 1: Added `commonLogger` to avoid having any dependency when using the logger during the first part of a framework lifecylce

edge33 (approved):
LGTM

packages/messages/migrations/sqls/20240708125718-users-from-uuid-to-nanoid-up.sql
```diff
@@ -3,53 +3,45 @@ BEGIN;
 ALTER TABLE users_imports
```
nnorbert, line 3: Shouldn't it be better to create a new migration? Existing migrations shouldn't be changed ever. Right now, it is not crucial because we can rebuild the whole DB, but in the future, this won't be possible.

nnorbert (approved):
LGTM!

PR #689, opened 2024-07-15
feat(19225): change formsie links

nnorbert:
change formsie links


alfonsograziano (commented):
Searching for `www.formsg.testing.` inside the codebase I can find multiple files affected and even an environment variable having that value (namely `NEXT_PUBLIC_FORMS_URL`). Is changing only this file enough for the scope of this PR?

alfonsograziano (approved):
LGTM

PR #692, opened 2024-07-16
[17330] Fix favicon for payments app

peschina:
### Ticket:

- [17330](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/17330)

### Description

Fix favicon for Payments app

## Type

- [ ] **Dependency upgrade**
- [x] **Bug fix**
- [ ] **New feature**
- [ ] **Dev change**

### Checklist:

### Screenshots:

N/A

### DIFF
--- a/apps/payments/app/[locale]/info/page.tsx
+++ b/apps/payments/app/[locale]/info/page.tsx
@@ -6,6 +6,7 @@ import Footer from "./Footer";
 import { getTranslations } from "next-intl/server";
 import Image from "next/image";
 import hero from "../../../public/landingPage/hero.png";
+import favicon from "../../../public/favicon.ico";
 import integration from "../../../public/landingPage/integration.png";
 import interfaceCitizen from "../../../public/landingPage/interfaceCitizen.png";
 import interfacePublicServant from "../../../public/landingPage/interfacePublicServant.png";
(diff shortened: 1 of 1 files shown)


apps/payments/app/[locale]/info/page.tsx
```diff
@@ -6,6 +6,7 @@ import Footer from "./Footer";
 import { getTranslations } from "next-intl/server";
 import Image from "next/image";
 import hero from "../../../public/landingPage/hero.png";
+import favicon from "../../../public/favicon.ico";
```
Comment (alfonsograziano) [inline apps/payments/app/[locale]/info/page.tsx:9]: Shall we do the same for all the other applications or just payments?

Comment (peschina) [reply apps/payments/app/[locale]/info/page.tsx:9]: The idea is to ensure it works for payments, then do the others in another PR. I need to do all the applications as part of the task :)

Comment (alfonsograziano) [reply apps/payments/app/[locale]/info/page.tsx:9]: Also, I can see that this fix has been applied to the landing page. Does it work across all the pages or just on this one? The landing pages are higher priority since we have actual traffic, but would be cool to find a solution what works across the entire app

alfonsograziano (approved):

apps/payments/app/[locale]/info/page.tsx
Comment (peschina) [reply apps/payments/app/[locale]/info/page.tsx:9]: I'll take care of the rest once we're sure this approach works

PR #711, opened 2024-07-19
Feature(profile-api): 15380 - add preferred language field

SamSalvatico:
### Ticket:

### Description

Added the preferred language field in user details

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**


apps/profile/migrations/sqls/20240719064612-add-preferred-language-to-user-details-up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE user_details
+ADD COLUMN preferred_language varchar(2) DEFAULT 'en'
```
Thurfjell, line 2: Hmm, this is fine for iso 639-1 but what if (big what if here) we need more bytes? en-gb or like 3 byte langs. 
Probably not but I don't really see the argument against going just plain old "text" here.

apps/profile-api/types/schemaDefinitions.ts
```diff
@@ -132,12 +132,14 @@ export const UpdateUserSchema = Type.Object({
   gender: Type.String(),
   phone: Type.String(),
   consentToPrefillData: Type.Optional(Type.Boolean()),
+  preferredLanguage: Type.Optional(Type.String()),
```
Thurfjell, line 135: I just assume if this is undefined, the field is not updated? I think this should be an integral part of the user, since it has default value in database?

SamSalvatico: If not set is not updated, yes

apps/profile/migrations/sqls/20240719064612-add-preferred-language-to-user-details-up.sql
SamSalvatico: I thought about it, but then I decided to do like this because will be pretty easy, if needed, in future, to cast to text

apps/profile-api/types/schemaDefinitions.ts
Thurfjell: What I meant is that I don't think it should be optional. Apologies for unclear.

SamSalvatico: I think should be, because If it is not set in the payload it won't be updated on the db, then the user will maintain the previous lang

SamSalvatico: I'll maintain this as mandatory here and optional in the patch one

Thurfjell (approved):

Thurfjell (approved):

PR #712, opened 2024-07-19
Chore(profile-api): use life events errors

SamSalvatico:
### Description

Use the LifeEvents errors as in other services

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [ ] **New feature**
- [x] **Dev change**

### Checklist:

### Screenshots:

N/A


apps/profile-api/routes/addresses/index.ts
```diff
@@ -51,7 +53,7 @@ export default async function addresses(app: FastifyInstance) {
 
         reply.send(result.rows);
       } catch (error) {
-        throw app.httpErrors.internalServerError((error as Error).message);
+        throw new ServerError(ERROR_PROCESS, (error as Error).message);
```
Thurfjell, line 56: Can we use a guard to check that its en error type instead?

Thurfjell (approved):

PR #713, opened 2024-07-19
fix(logto) Added safety checks for logto users

edge33:
### Description

fixes issues with edge cases in logto:

- crash when removing an user from logto while the said user is logged into life-events
- crash when adding roles to user when the said user is logged in life-events

## Type

- [ ] **Dependency upgrade**
- [x] **Bug fix**
- [ ] **New feature**
- [ ] **Dev change**

### Checklist:

### Screenshots:

N/A

### DIFF
--- a/packages/auth/src/auth-session.ts
+++ b/packages/auth/src/auth-session.ts
@@ -15,6 +15,7 @@ import { redirect } from "next/navigation";
 import { LogtoNextConfig, UserScope } from "@logto/next";
 import { BadRequestError } from "shared-errors";
 import { decodeJwt } from "jose";
+import { getCommonLogger } from "nextjs-logging-wrapper";
 
 const PROCESS_ERROR = "PARSE_LOGTO_CONTEXT";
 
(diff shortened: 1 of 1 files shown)


packages/auth/src/auth-session.ts
```diff
@@ -46,16 +46,25 @@ export const AuthSession: IAuthSession = {
...
     }
-
     addInactivePublicServantScope(config);
-
-    const context = await getLogtoContext(config, getContextParameters);
+    let context;
+    try {
+      context = await getLogtoContext(config, getContextParameters);
+    } catch (err) {
+      console.error(err);
```
SamSalvatico, line 54: do not use console log
you can check how to log in auth package here: 
https://github.com/ogcio/life-events/blob/abb6499f6fb4eb4b6c3d88c60bf2ef366100702e/packages/auth/src/base-authentication-context.ts#L76

SamSalvatico (approved):

PR #714, opened 2024-07-19
16375 payment request entity

nnorbert:
Refactored payments request into an entity service.

### DIFF
--- a/apps/payments-api/plugins/entities/paymentRequest/index.ts
+++ b/apps/payments-api/plugins/entities/paymentRequest/index.ts
@@ -180,6 +180,31 @@ const buildUpdatePaymentRequest =
     }
   };
 
+const buildDeletePaymentRequest =
+  (repo: PaymentRequestRepo, httpErrors: HttpErrors) =>
+  async (paymentRequestId: string, organizationId: string) => {
+    const transaction = repo.getTransaction();
+    try {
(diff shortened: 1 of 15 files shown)


apps/payments-api/plugins/entities/paymentRequest/index.ts
```diff
@@ -0,0 +1,241 @@
...
+    try {
+      result = await repo.getPaymentRequestsTotalCount(organizationId);
+    } catch (err) {
+      log.error((err as Error).message);
+    }
+
+    const totalCount = result?.rows[0].totalCount;
+
+    if (totalCount === undefined) {
+      throw httpErrors.internalServerError("Something went wrong.");
```
peschina, line 51: Do we want to use errors from the shared-errors packages as part of this PR?

peschina (approved):
Looks good.
I think we should discuss about the errors, but doesn't necessarily needs to be part of this PR

apps/payments-api/plugins/entities/paymentRequest/index.ts
nnorbert: I don't really want to add too much to this PR. This is only about refactoring, so the errors were just moved, but they are the same. I recommend a specific PR for error handling.

PR #716, opened 2024-07-22
Feature(messaging): prepare language selector

SamSalvatico:
### Ticket:

### Description

Use the same header as other services do

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**


apps/messaging/app/[locale]/LanguageSwitch.tsx
```diff
@@ -0,0 +1,91 @@
...
+  light: ds.colours.ogcio.darkGreen,
+};
+
+export default async ({ theme }: { theme: Theme }) => {
+  const lang = headers().get("x-next-intl-locale");
+
+  const handleLanguageChange = async (lang: string) => {
+    "use server";
+
+    const referer = headers().get("referer");
```
Comment (Thurfjell) [inline apps/messaging/app/[locale]/LanguageSwitch.tsx:19]: Are you sure this is the correct header? This isn't super obvious in next with the Link eg. 
We have an x-pathname or something which in my experience yield expected result.

I might be totally wrong here, just raising the concern.

Comment (SamSalvatico) [reply apps/messaging/app/[locale]/LanguageSwitch.tsx:19]: I need the help of the super @nnorbert to answer this , I barely copied his code here 🆘 🙇

Comment (SamSalvatico) [reply apps/messaging/app/[locale]/LanguageSwitch.tsx:19]: @nnorbert can you help use here?

Comment (nnorbert) [reply apps/messaging/app/[locale]/LanguageSwitch.tsx:19]: The x-pathname was a custom header, I remember that one. I also checked the links section from the headers out, but I stuck with the referer, which is a standard header and tells you the path from the user was coming. Referer should work just fine since it is a standard header [https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referer](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referer)

Comment (Thurfjell) [reply apps/messaging/app/[locale]/LanguageSwitch.tsx:19]: Yep, that's what's expected. I just had some issues with this header not being what I thought it would. Let's go with this and deal with any bugs

Thurfjell (dismissed):

Thurfjell (dismissed):

Thurfjell (approved):

PR #717, opened 2024-07-22
Feature(messaging): 15380 - preferred language

SamSalvatico:
### Ticket:

### Description

Set the preferred language and read it from the DB

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**


SamSalvatico:
Why I am adding everybody here?
Because at the moment I implemented this for the messaging, but in future we will use a shared language across building blocks, using the one stored in the user profile

packages/building-blocks-sdk/services/profile/index.ts
```diff
@@ -21,7 +21,7 @@ export class Profile {
         req.headers.set("x-user-id", authToken);
```
SamSalvatico, line 21: We still have to integrate the user profile with Logto, so I am maintaning this, will update in the next pr

nnorbert (approved):
LGTM!

PR #725, opened 2024-07-24
[Payments] Organization selector in homepage only

peschina:
### Ticket:

- [19298](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/19298)

### Description

Display organization selector component only in homepage. To keep `PaymentsMenu` component as SSR and render the organization selector dynamically based on route we render it once per route. 

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**

### Checklist:

### Screenshots:

N/A


apps/payments/app/[locale]/(hosted)/paymentSetup/PaymentsMenu.tsx
```diff
@@ -35,18 +37,20 @@ export default ({
...
-            organizations={organizations.map((org) => ({
-              name: org.name,
-              id: org.id,
-            }))}
-            defaultOrganization={defaultOrganization}
-            handleSubmit={handleSubmit}
-          ></OrganizationSelector>
-        )}
+        <Suspense fallback={<h3>{t("loading")}</h3>}>
+          {organizations && organizations.length > 1 && (
```
Comment (nnorbert) [inline apps/payments/app/[locale]/(hosted)/paymentSetup/PaymentsMenu.tsx:41]: This check can be simplified to organisations?.length > 1

apps/payments/app/[locale]/(hosted)/paymentSetup/page.tsx
```diff
@@ -42,97 +49,112 @@ export default async function ({
     return redirect("/error", RedirectType.replace);
   }
 
+  const locale = headers().get("x-next-intl-locale") || "";
```
Comment (nnorbert) [inline apps/payments/app/[locale]/(hosted)/paymentSetup/page.tsx:52]: This is a page, can't the locale be accessed from the query params?

Comment (peschina) [reply apps/payments/app/[locale]/(hosted)/paymentSetup/page.tsx:52]: You're right!

apps/payments/app/[locale]/(hosted)/paymentSetup/PaymentsMenu.tsx
Comment (peschina) [reply apps/payments/app/[locale]/(hosted)/paymentSetup/PaymentsMenu.tsx:41]: I think with `organisations?.length > 1` ts will complain. it will also complain for line 45

Comment (nnorbert) [reply apps/payments/app/[locale]/(hosted)/paymentSetup/PaymentsMenu.tsx:41]: Probably because you are comparing against a number, and undefined will cause an issue there. I see. Okay, thanks!

nnorbert (approved):
LGTM!

PR #730, opened 2024-07-24
feat(upload-api) Implemented better download logic

edge33:
### Description

Implements a better control of the infected file download

## Type

- [ ] **Dependency upgrade**
- [x] **Bug fix**
- [ ] **New feature**
- [x] **Dev change**

### Checklist:

### Screenshots:

N/A


apps/upload-api/src/routes/files/index.ts
```diff
@@ -335,41 +330,30 @@ export default async function routes(app: FastifyInstance) {
...
-        //   this.push(this.lastChunk);
-        // } else {
-        //   this.push();
-        // }
+        this.push(this.lastChunk);
         callback();
       })
-      .catch(callback);
+      .catch((err) => {
+        this.emit("error", new CustomError(FILE_DOWNLOAD, err, 400));
```
SamSalvatico, line 355: Why don't directly use `BadRequestError` if code is 400?
Which are the possible causes of this error?
I mean, is a 4xx or 5xx?

apps/upload-api/src/routes/files/index.ts
```diff
@@ -335,41 +330,30 @@ export default async function routes(app: FastifyInstance) {
 
 class PromiseTransform extends Transform {
   private aPromise;
-  // private lastChunk;
+  private lastChunk: Buffer | undefined;
```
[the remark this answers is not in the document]
edge33: it is a way we use to control the download flow, the file is processed chunk by chunk, so we store in memory the `lastChunk` the `transform` gets. when the processing is done we only flush the stream out when the promise resolve, and keeping this lastChunk out we can decide to stream the very last chunk to close the file or to emit an error and make the server respond with an error

apps/upload-api/src/routes/files/index.ts
edge33: this happens when a file is marked as infected with a virus, I am not sure it is a 500, because it is not en error with the app itself but rather an expected behaviour with wrong/unexpected data. what do you think should be the error code in this case?

edge33 (commented):
Replied to comments

Thurfjell (approved):

SamSalvatico (approved):

PR #731, opened 2024-07-24
Chore(shared-errors): added parent error management

SamSalvatico:
### Description

I've added a `parentError` field in the shared errors so we can avoid to log the "original" error prior to throw a Life Events one

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**


edge33 (approved):
LGTM

PR #737, opened 2024-07-26
Provider API refactor

Thurfjell:
### Ticket:

- [19955](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/19955)

### Description

Refactor all provider API endpoints (sms and email) into a `/providers` resource.
Removed all old references.

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [ ] **New feature**
- [x] **Dev change**

### Checklist:

### Screenshots:

N/A


apps/messaging-api/routes/providers/index.ts
```diff
@@ -0,0 +1,722 @@
...
+      const totalCount = query.rows.at(0)?.count || 0;
+      const url = new URL(`/api/v1/${prefix}`, process.env.HOST_URL).href;
+      const links = getPaginationLinks({
+        totalCount,
+        url,
+        limit,
+        offset,
+      });
+
+      const response: GenericResponseSingle<Static<typeof ProviderList>> = {
```
SamSalvatico, line 220: Why using "Single" type given that the response is a list?

apps/messaging/app/[locale]/admin/providers/SmsProviders.tsx
```diff
@@ -14,16 +14,14 @@ export default async () => {
 
   const { data: providers } = await sdk.getSmsProviders();
 
+  console.log(providers);
```
Comment (SamSalvatico) [inline apps/messaging/app/[locale]/admin/providers/SmsProviders.tsx:17]:
```suggestion

```

apps/messaging-api/routes/providers/index.ts
Thurfjell: Aha I think we should actually get rid of the one setting the generic type to an array. That's definitely something that the implementor should have to decide imo.

Some bad naming.

SamSalvatico: agree

PR #749, opened 2024-07-30
feat: update payment dependencies

nnorbert:
Update payment dependencies


SamSalvatico:
Do we still need this?

PR #752, opened 2024-07-31
[Payments] Update Logto sdk version

peschina:
### Ticket:

- [19880](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/19880)

### Description

Update Logto sdk version to use the new `getAccessToken` and `getOrganizationToken` methods and so avoid getting a new token at every API call.
The access token is now decoupled from the context, and can be correctly cached only if retrieved within a server action or a route handler called from the browser.
Atm we're missing a proper solution for when we need to call a route handler from within a RSC, the temporary workaround is to load the token with an hidden image so that it can be cached.

## Type

- [ ] **Dependency upgrade**
- [x] **Bug fix**
- [ ] **New feature**
- [x] **Dev change**

### Checklist:

### Screenshots:

N/A


SamSalvatico (approved):
Great on the messaging side

apps/messaging/app/api/token/route.ts
```diff
@@ -0,0 +1,8 @@
+import { NextResponse } from "next/server";
+import { AuthenticationFactory } from "../../utils/authentication-factory";
+
+export async function GET() {
+  const token = await AuthenticationFactory.getInstance().getToken();
```
alfonsograziano, line 5: Since the solution we're adopting is non-trivial, shall we write a few lines of documentation explaining why we need this route here?

apps/messaging/app/utils/authentication-factory.ts
```diff
@@ -1,41 +1,34 @@
...
-      return new Messaging(await params.authenticationContext.getAccessToken());
-    }
+  static async getToken(): Promise<string> {
+    const cookieHeader = headers().get("cookie") as string;
 
-    const token = await this.getInstance().getAccessToken();
+    const res = await fetch(
+      new URL(
+        "/api/token",
+        process.env.NEXT_PUBLIC_MESSAGING_SERVICE_ENTRY_POINT as string,
```
alfonsograziano, line 17: Same here, do you think is worth explaining why we're calling the endpoint, why we're forwarding the cookies to the fetch etc?

I think the solution makes sense, but a new developer would think "why aren't we just calling the method?"

apps/messaging/app/utils/authentication-factory.ts
```diff
@@ -1,41 +1,34 @@
...
+      ),
+      { headers: { cookie: cookieHeader } },
+    );
+    const { token } = await res.json();
+    return token;
+  }
 
+  static async getMessagingClient(): Promise<Messaging> {
```
alfonsograziano, line 25: non-blocking: shall we consider creating a getPaymentsClient as well?

apps/messaging/package.json
```diff
@@ -17,7 +17,7 @@
     "exec": "next dev -p 3002"
   },
   "dependencies": {
-    "@logto/next": "^3.3.0",
+    "@logto/next": "^3.4.0",
```
alfonsograziano, line 20: non-blocking: are we really using this package inside of the messaging application? The main idea behind a shared package is to move the logto dependency outside of the single app

apps/payments/app/[locale]/(hosted)/layout.tsx
```diff
@@ -25,6 +25,11 @@ export default async function RootLayout({
           flexDirection: "column",
         }}
       >
+        {/* workaround to allow Logto to cache the access token */}
+        <img
```
Comment (alfonsograziano) [inline apps/payments/app/[locale]/(hosted)/layout.tsx:29]: What if we create a component for it like `LogtoTokenRefresher` and in the component file we explain exactly what we're doing, why it works and what are the implications?

I know this is a temporary solution, but in a few weeks we might forget all the details we currently have in mind

apps/profile/app/utils/authentication-factory.ts
```diff
@@ -1,24 +1,21 @@
...
-      return new Profile(params.token);
-    }
-    if (params?.authenticationContext) {
-      return new Profile(await params.authenticationContext.getAccessToken());
-    }
+  static async getProfileClient(): Promise<Profile> {
+    const cookieHeader = headers().get("cookie") as string;
 
-    const token = await this.getInstance().getAccessToken();
+    const res = await fetch(
```
alfonsograziano, line 14: non-blocking: in case we notice that this solution is the one we we will use in the long-term, maybe is worth exposing this method (with a configurable /token endpoint) as part of the `auth` package instead of duplicating code. In this way we can also document in all our rationales in one place

alfonsograziano (dismissed):
LGTM, it might be worth explaining a little bit the solution we proposed and what are the implications. This can be done in another PR as well

apps/messaging/app/api/token/route.ts
edge33: AFAIU this is done to allow easier caching of the token

edge33 (approved):
LGTM

apps/profile/app/utils/authentication-factory.ts
peschina: Sure, I would do this in a separate PR in the near future, say as soon as we hear back from Logto people

apps/messaging/package.json
peschina: We use it in all applications to import the handleSignIn method, but we actually don't need it in the dependencies. [aea445a](https://github.com/ogcio/life-events/pull/752/commits/aea445a072160e85f957d7a0e3359ff285721ece)

apps/payments/app/[locale]/(hosted)/layout.tsx
Comment (peschina) [reply apps/payments/app/[locale]/(hosted)/layout.tsx:31]: [2093b43](https://github.com/ogcio/life-events/pull/752/commits/2093b4346fb51f515a8e876a98c0122e03057f00)

apps/messaging/app/utils/authentication-factory.ts
peschina: [3cb6b37](https://github.com/ogcio/life-events/pull/752/commits/3cb6b37526050bddfd18d079b9b1c4d5f01345b7)

apps/messaging/app/api/token/route.ts
peschina: [3cb6b37](https://github.com/ogcio/life-events/pull/752/commits/3cb6b37526050bddfd18d079b9b1c4d5f01345b7)

alfonsograziano (approved):

PR #760, opened 2024-08-01
Chore(messaging-api): single entry point for process messages

SamSalvatico:
### Description
Now we use the process messages method instead of rewriting the code to create and schedule messages

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [ ] **New feature**
- [x] **Dev change**


Thurfjell (approved):
LGTM 🥹

PR #762, opened 2024-08-02
[Payments] Fix login in E2E tests locally run

peschina:
### Ticket:

- [20090](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/20090)

### Description

Fix login fixture in E2E tests for Payments. The assumption is that tests are run locally, so we're testing the integration with Logto and our MyGovId mock service

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [ ] **New feature**
- [] **Dev change**
- [x] **Additional tests**

### Checklist:

### Screenshots:

N/A


apps/payments/playwright.config.ts
```diff
@@ -1,52 +1,43 @@
...
-  timeout: 30 * 1000,
+  timeout: 500000,
+  grep: testPlanFilter(),
+  reporter: [["line"], ["allure-playwright"], ["html"]], // Test directory
   testDir: path.join(__dirname, "e2e"),
-  retries: process.env.CI ? 1 : 0,
+  retries: 0,
   outputDir: "./e2e/test-results/",
-  workers: process.env.CI ? 1 : undefined,
+
```
peschina, line 19: This can be uncommented at any time and will spin up our services and dependencies everytime a test command is run

apps/payments/e2e/fixtures/test.ts
```diff
@@ -1,9 +1,13 @@
 import { test as baseTest, expect } from "@playwright/test";
 import fs from "fs";
 import path from "path";
+import { LoginPage } from "../objects/loginPage";
+import { password, publicServantUser } from "../utils/constants";
 
-const baseURL = process.env.HOST_URL ?? "http://localhost:3001";
-const loginUrl = process.env.LOGIN_URL ?? "http://localhost:8000/static/login/";
+const baseURL =
```
alfonsograziano, line 7: non-blocking: do we know if there is any way to define the env vars we need with playwright with a json schema like in Fastify? I'd personally prefer our tests to fail instead of having a fallback that might be wrong depending on the environment

apps/payments/e2e/objects/loginPage.ts
```diff
@@ -0,0 +1,32 @@
+import { Locator, Page, expect } from "@playwright/test";
+import { publicServantEmailDomain } from "../utils/constants";
+
+export class LoginPage {
```
alfonsograziano, line 4: non-blocking: just as a suggestion, it might be renamed to `MyGovIdMockLoginPage` since in the future we will have EntraID and possibly the real MyGovId integration

apps/payments/e2e/utils/constants.ts
```diff
@@ -1,3 +1,6 @@
 export const paymentSetupUrl = "/en/paymentSetup";
 export const paymentRequestUrl = `${paymentSetupUrl}/requests`;
 export const providersUrl = `${paymentSetupUrl}/providers`;
+export const publicServantUser = "Tony Stark";
```
alfonsograziano, line 4: non-blocking: Might this be an object instead? Like `myGovIdMockSettings`

alfonsograziano (dismissed):
LGTM

apps/payments/e2e/fixtures/test.ts
peschina: As far as I can see the env vars are injected if you enable the devServer config. If you run the application you want to test in a separate process, as I'm doing now, you have no env vars at all. It's something that we can further investigate in a separate PR :)

alfonsograziano (approved):

rgrnearform:
@peschina Nothing to do with this PR, the only thing I would change in the payments is to store the spec files in a spec folder, at least have this in common so will be easier to William to setting up in the CI, instead of having microservices having different structures regarding where the spec files are stored.

rgrnearform (approved):
Left a comment but nothing related to this PR. LGTM

PR #775, opened 2024-08-05
Fix(messaging-api): missing recipient and mandatory organization id

SamSalvatico:
### Description

Added a check to ensure organizationId is set when needed

## Type

- [ ] **Dependency upgrade**
- [x] **Bug fix**
- [ ] **New feature**
- [ ] **Dev change**


edge33 (approved):
LGTM

PR #777, opened 2024-08-05
Feature(messaging-api): improve get message response

SamSalvatico:
### Description

Added missing fields in the get single message response

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**


apps/messaging-api/services/messages/messages.ts
```diff
@@ -43,10 +43,15 @@ export const getMessage = async (params: {
   const data = await params.pg.query<ReadMessage>(
     `
         select 
-            subject, 
-            excerpt, 
+            subject as "subject",
```
alfonsograziano, line 46: non-blocking: is this `as` really needed?

alfonsograziano (approved):

apps/messaging-api/services/messages/messages.ts
SamSalvatico: It's just a consequence of my wanting to standardize everything; as you say, it wouldn't be necessary at all 😸

PR #788, opened 2024-08-06
Fix isAuthenticated method

alfonsograziano:
The old isAuthenticated method was requesting the AuthSession.get (which redirects to login if you're not authenticated)


SamSalvatico (approved):

alfonsograziano:
Closing, since the changes has been merged already by #782

PR #792, opened 2024-08-07
[20241] Reduce payments api image size

peschina:
### Ticket:

- [20241](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/20241)

### Description

increase docker image size from 1.82GB to 738MB.
Use production build.

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [ ] **New feature**
- [x] **Dev change**

### Checklist:

### Screenshots:

N/A


apps/payments-api/Dockerfile
```diff
@@ -2,30 +2,60 @@ FROM base-deps AS deps
...
+FROM deps AS runner
+
+COPY --from=assembler /app /app
+WORKDIR /app/apps/payments-api
 
 ENV NODE_ENV=production
 ENV LOG_LEVEL=trace
 EXPOSE 8001
-CMD [ "npm", "--prefix", "apps/payments-api",  "run", "start" ]
+CMD [ "npx", "ts-node", "dist/", "index.js" ]
```
alfonsograziano, line 61: I see you're creating a build command and a different start. Any reason to run `npx ts-node` here instead of the `npm run start`?

apps/payments-api/services/realexService.ts
```diff
@@ -1,11 +1,11 @@
-import { RealexHppResponse } from "../routes/schemas";
+import { RealexHppResponseDO } from "../plugins/entities/providers/types";
```
alfonsograziano, line 1: non-blocking: at a first look, this seems uncorrelated with this PR but the change itself is reasonable

alfonsograziano (approved):
LGTM

apps/payments-api/services/realexService.ts
peschina: I had to fix it otherwise the build would fail :)

PR #805, opened 2024-08-08
feat(upload) Allows upload/download/deletion of files with metadata

edge33:
### Ticket:

- [18787](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/18787)

### Description

- Allows download/upload/list of all files in the upload app

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**
- [ ] **Additional tests**
- [ ] **Documentation**
- [ ] **Other**

### Checklist:

### Screenshots:

N/A


apps/upload-api/db/migrations/sqls/20240808072457-add-filename-field-up.sql
```diff
@@ -0,0 +1,9 @@
+BEGIN;
+-- Add the filename column with a default value of NULL
+ALTER TABLE files ADD COLUMN filename VARCHAR(255) DEFAULT NULL;
+
+-- Update existing records to set filename to the extracted value from the key field
+UPDATE files
+SET filename = substring(key from '([^/]+)$');
+
+COMMIT;
```
alfonsograziano, line 9: EOF :) - we should add something in our config to automate this on sql files as well XD

apps/upload-api/src/routes/files/index.ts
```diff
@@ -23,10 +18,13 @@ import {
...
   ServerError,
 } from "shared-errors";
+import deleteFileMetadata from "./utils/deleteFileMetadata.js";
+import getFileMetadata from "./utils/getFileMetadata.js";
+import getUserFiles from "./utils/getUserFiles.js";
 import insertFileMetadata from "./utils/insertFileMetadata.js";
 
 const FILE_UPLOAD = "FILE_UPLOAD";
-const FILE_INDEX = "FILE_INDEX";
+// const FILE_INDEX = "FILE_INDEX";
```
alfonsograziano, line 27: In case you don't need this var anymore is worth keeping it with a comment?

apps/upload-api/src/routes/files/index.ts
```diff
@@ -411,13 +407,24 @@ export default async function routes(app: FastifyInstance) {
       if (!request.params.key) {
         throw new BadRequestError(FILE_DELETE, "File key not provided");
       }
+
+      const fileData = await getFileMetadata(app, request.params.key, userId);
+
+      const file = fileData.rows.length > 0 ? fileData.rows[0] : undefined;
```
alfonsograziano, line 413: non-blocking: `fileData.rows?.[0]` should do the same

apps/upload-api/src/routes/files/index.ts
```diff
@@ -462,17 +481,25 @@ export default async function routes(app: FastifyInstance) {
...
       const antivirusPassthrough = app.avClient.passthrough();
       const downloadPassthrough = new PassThrough();
 
       const thePromise = new Promise<void>((resolve, reject) => {
         antivirusPassthrough.once("error", (err) => {
-          return reject(
-            new ServerError(FILE_DOWNLOAD, "Internal server error", err),
-          );
+          app.log.error(err);
+          // Rejecting here can cause html 5 video to cause ERR_STREAM_PREMATURE_CLOSE
```
alfonsograziano, line 492: non-blocking: do we need to keep this?

apps/upload-api/src/routes/files/utils/deleteFileMetadata.ts
```diff
@@ -0,0 +1,19 @@
...
+  fileSize: number;
+  mimetype: string;
+  createdAt: Date;
+  lastScan: Date;
+  infected: boolean;
+  infectionDescription: string | null;
+};
+
+export default async (app: FastifyInstance, fileId: string) => {
+  return await app.pg.query<FileMetadata>(`DELETE FROM files WHERE ID = $1`, [
```
alfonsograziano, line 16: non-blocking: that await is not necessary

apps/upload-api/src/routes/files/utils/getFileMetadata.ts
```diff
@@ -0,0 +1,22 @@
...
+  fileSize: number;
+  mimetype: string;
+  createdAt: Date;
+  lastScan: Date;
+  infected: boolean;
+  infectionDescription: string | null;
+};
+
+export default async (app: FastifyInstance, fileId: string, owner: string) => {
+  return await app.pg.query<FileMetadata>(
```
alfonsograziano, line 17: same

apps/upload/app/[locale]/(public)/components/FileTable.tsx
```diff
@@ -0,0 +1,77 @@
+import DeleteFile from "./DeleteFile";
+
+const formatBytes = (bytes: number) => {
```
Comment (alfonsograziano) [inline apps/upload/app/[locale]/(public)/components/FileTable.tsx:3]: this sounds like a magical utility. Did u copied it from some GH repo or so? In case, it might be helpful to add a reference with a comment just for future reference

apps/upload/app/[locale]/(public)/components/FileTable.tsx
```diff
@@ -0,0 +1,77 @@
...
+  if (bytes === 0) return "0 Bytes";
+  const k = 1000;
```
Comment (alfonsograziano) [inline apps/upload/app/[locale]/(public)/components/FileTable.tsx:5]: I do't 100% get what this code does, I guess converts raw bytes in a readable manner. In case, just a small comment, `k` might be 1024

apps/upload/app/api/file/[fileId]/route.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export async function GET(
+  request: Request,
+  { params: { fileId } }: { params: { fileId: string } },
+) {
+  await AuthenticationFactory.getInstance().getContext();
+  const uploadClient = await AuthenticationFactory.getUploadClient();
+  const { data, error, headers, status } = await uploadClient.getFile(fileId);
+
+  if (error || !data) {
```
Comment (alfonsograziano) [inline apps/upload/app/api/file/[fileId]/route.ts:13]: non-blocking: the logic here is not 100% clear. I had to read it a couple of times (mainly due to a doubled very similar condition). There might be some room for refactoring in case

alfonsograziano (dismissed):
Just included a few optional comments about things that might be improved or clarified, but overall is awesome :D 

LGTM

apps/upload-api/src/routes/files/index.ts
edge33: I will need to discuss this with Tiago, but for now I would keep it as a reminder that we have this edge case

apps/upload/app/[locale]/(public)/components/FileTable.tsx
Comment (edge33) [reply apps/upload/app/[locale]/(public)/components/FileTable.tsx:5]: It's just some code that copilot made, I am open to suggestion here, I needed a quick and dirty temporary solution.

for the sizing, `k` would be 1024 if we want to display the unit in Mebibytes, while 1000 is for Megabytes, I don't have a strict preference here. I used 1000 in this case to match the file size of my files on MacOs Finder.

apps/upload/app/api/file/[fileId]/route.ts
Comment (edge33) [reply apps/upload/app/api/file/[fileId]/route.ts:13]: happy to apply a change if you envisage a different approach :)

apps/upload-api/src/routes/files/index.ts
alfonsograziano: Fair enough!

apps/upload/app/[locale]/(public)/components/FileTable.tsx
Comment (alfonsograziano) [reply apps/upload/app/[locale]/(public)/components/FileTable.tsx:5]: Sounds good :)

apps/upload/app/api/file/[fileId]/route.ts
Comment (alfonsograziano) [reply apps/upload/app/api/file/[fileId]/route.ts:13]: no worries, it make sense after reading it a couple of times :)

PR #813, opened 2024-08-09
[Payments] E2E test - Inactive payment request creation with manual bank transfer provider

peschina:
### Ticket:

- [20092](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/20092)

### Description

Test creation of inactive payment request with a manual bank transfer provider.
Tidy up provider validation error types.

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [ ] **New feature**
- [ ] **Dev change**
- [ ] **Additional tests**
- [x] **Documentation**
- [ ] **Other**

### Checklist:

### Screenshots:

N/A


apps/payments/e2e/objects/paymentRequests/PaymentRequestDetailsPage.ts
```diff
@@ -0,0 +1,148 @@
...
+      "Allow amount override from url",
+    );
+    this.customAmountLabel = this.page.getByText(
+      "Allow the user to pay a custom amount",
+    );
+    this.paymentLinkLabel = this.page.getByText("Payment link");
+    this.paymentLink = this.page
+      .locator("div")
+      .filter({ hasText: "Payment link" })
+      .last()
```
alfonsograziano, line 43: non-blocker: just for my benefit, why is it necessary to use the `.last()` in some places?

apps/payments/e2e/objects/paymentRequests/PaymentRequestDetailsPage.ts
```diff
@@ -0,0 +1,148 @@
...
+    ).toBeVisible();
+  }
+
+  async delete() {
+    await this.deleteButton.click();
+  }
+
+  async getPaymentLink() {
+    await expect(this.paymentLinkLabel).toBeVisible();
+    return (await this.paymentLink.textContent()) ?? "";
```
alfonsograziano, line 146: Is `?? ""` really necessary? Shall we expect that `paymentLink.textContent()` is not empty and then cast as string if necessary? I'd prefer throwing an error rather than returning an empty string if the payment link is not generated correcly

alfonsograziano (approved):
LGTM

apps/payments/e2e/objects/paymentRequests/PaymentRequestDetailsPage.ts
peschina: In this case we can't target the label by an aria role. So we're getting all the divs that contain that text, and this means also getting all the divs that have that text in one of their children. We need to get the last div (with `.last()`) so we can target the most inner one - in our case the one that has a link in it

apps/payments/e2e/objects/paymentRequests/PaymentRequestDetailsPage.ts
peschina: We expect the payment link to always be defined, but `locator.textContent()` returns string or undefined by signature. There is usually no need to throw custom error in e2e tests as assertions already cover that. If the link is empty the test will fail at the following assertion, when trying to open a page from an empty link :)

PR #835, opened 2024-08-14
[Payments] E2E - test transaction with manual bank transfer

peschina:
### Ticket:

- [20083](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/20083)

### Description

Test initiation of a manual bank transfer transaction.
Added another worker scoped fixture to log in a citizen, so we can two different auth files, one with the public servant auth cookies and one with the citizen auth cookies. We'll use pages with an authenticated public servant by default. 
Entrypoint pages with a citizen logged in will be available via fixture - we can discuss exposing just a generic page fixture with context and storage state dedicated to citizen without instantiating a page class if you guys prefer. 

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [ ] **New feature**
- [ ] **Dev change**
- [x] **Additional tests**
- [ ] **Documentation**
- [ ] **Other**

### Checklist:

### Screenshots:

N/A


apps/payments/e2e/fixtures/auth.ts
```diff
@@ -10,15 +10,19 @@ const baseURL =
...
 
 export * from "@playwright/test";
-export const test = baseTest.extend<{}, { workerStorageState: string }>({
-  storageState: ({ workerStorageState }, use) => use(workerStorageState),
+export const test = baseTest.extend<
+  {},
+  { pubServantWorkerStorageState: string; userWorkerStorageState: string }
+>({
+  storageState: ({ pubServantWorkerStorageState }, use) =>
+    use(pubServantWorkerStorageState),
```
peschina, line 18: we're using the storage state with the public servant cookies as default

apps/payments/e2e/fixtures/auth.ts
```diff
@@ -53,4 +57,43 @@ export const test = baseTest.extend<{}, { workerStorageState: string }>({
     },
     { scope: "worker" },
   ],
+
+  userWorkerStorageState: [
```
peschina, line 61: we write the auth cookies for citizen once per worker, so they're ready to be injected in the context when neeeded

apps/payments/e2e/fixtures/citizenPagesFixtures.ts
```diff
@@ -0,0 +1,17 @@
...
+};
+
+export const test = base.extend<citizenPagesFixtures>({
+  payPage: async ({ browser, userWorkerStorageState }, use) => {
+    const context = await browser.newContext({
+      storageState: userWorkerStorageState,
+    });
+    const payPage = new PayPage(await context.newPage());
+    await use(payPage);
+    await context.close();
```
peschina, line 15: When we want to test a page with a citizen logged in we need a different context that uses the storage state with the citizen auth cookies. We provide the entrypoint pages (as in this case with the pay page) as a test fixture because the files containing the auth cookies have a dynamic name with the worker id in it, so we can't hardcode it inside a test but we can access it as a worker scoped fixture inside a test fixture.

apps/payments/e2e/specs/paymentRequests/paymentRequestMultipleProviders.spec.ts
```diff
@@ -88,6 +88,7 @@ test.describe("Payment Request with multiple providers", () => {
     const inactivePayPage = new InactivePayPage(newPage);
     await inactivePayPage.checkHeader();
     await inactivePayPage.checkDescription();
+    await newPage.close();
```
peschina, line 91: best to always close new tabs when they are no longer needed

apps/payments/e2e/specs/transactions/manualBankTransfer.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    const publicServantPage = await browser.newPage();
+    const paymentRequestsPage = new PaymentRequestsPage(publicServantPage);
+    await paymentRequestsPage.goto();
+    await paymentRequestsPage.gotoDetails(paymentRequestWithMultipleProviders);
+
+    const detailsPage = new PaymentRequestDetailsPage(publicServantPage);
+    const paymentLink = await detailsPage.getPaymentLink();
+
+    const citizenPage = payPage.page;
```
peschina, line 41: This is the page fixture coming from our custom fixture. When we want to test pages with a citizen logged in we need to use this one

rgrnearform (approved):
I like it the way you are handling the auth.
Looks good to me. well done

PR #853, opened 2024-08-21
Messaging Providers API tests

Thurfjell:
### Ticket:

- [20299](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/20299)

### Description

Tests are set up to be dependent on each other (more or less) in a sequence. (set by meta field seq in .bru files).
Each folder would represent a suite that should run in sync.
Example
Providers/Emails <- all tests will run from top to bottom, and depend on the results of the previous (to some degree)
Providers/Sms would be independent to Providers/Email eg.

One other option would be to pre-request set up via API calls, whatever resources required by the main test, to make them independent. This would produce quite some overhead in calls and resources created.

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [ ] **New feature**
- [x] **Dev change**
- [x] **Additional tests**
- [ ] **Documentation**
- [ ] **Other**

### Checklist:

### Screenshots:

N/A


edge33 (approved):
LGTM

nnorbert (approved):
I am not an expert in Bruno tests, but it looks good!

PR #858, opened 2024-08-22
20084 e2e pay stripe

nnorbert:
E2E test payment with Stripe provider

### DIFF
--- a/apps/payments/e2e/specs/transactions/stipe.spec.ts
+++ b/apps/payments/e2e/specs/transactions/stipe.spec.ts
@@ -1,0 +1,57 @@
+import { test } from "../../fixtures/paymentRequestsFixtures";
+import {
+  Severity,
+  owner,
+  tags,
+  severity,
+  description,
+} from "allure-js-commons";
+import { PaymentRequestsPage } from "../../objects/paymentRequests/PaymentRequestsListPage";
+import { PaymentRequestDetailsPage } from "../../objects/paymentRequests/PaymentRequestDetailsPage";
+import {
+  mockAccountHolderName,
+  mockAmount,
+  mockRedirectUrl,
+} from "../../utils/mocks";
+import { TrueLayerDialogPage } from "../../objects/payments/openbanking/TrueLayerDialogPage";
+import { expect } from "@playwright/test";
+import { PayPage } from "../../objects/payments/PayPage";
+import { StripeForm } from "../../objects/payments/stripe/StripeForm";
+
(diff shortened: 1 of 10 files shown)


apps/payments/e2e/specs/transactions/stipe.spec.ts
rgrnearform, line 1: typo stipe.ts

rgrnearform (commented):
typo

apps/payments/e2e/specs/transactions/stipe.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+test.describe("Transaction with stripe", () => {
+  const cards = {
+    success: "4242 4242 4242 4242",
+    genericDecline: "4000 0000 0000 0002",
+    insufficientFunds: "4000 0000 0000 9995",
+    incorrectNumber: "4242 4242 4242 4241",
+  };
+
+  const securityCode = "123";
+  let expirationDate: string;
```
peschina, line 26: we could optionally just make it a const. Like a month in 2040 😄

apps/payments/e2e/specs/transactions/stipe.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+    await stripe.fillCardNumber(cards.success);
+    await stripe.fillExpDate(expirationDate);
+    await stripe.fillSecurityCode(securityCode);
+    await stripe.pay();
+
+    await expect(
+      citizenPage.getByRole("img", { name: "Google" }),
+    ).toBeVisible();
+    await expect(citizenPage.url()).toContain(mockRedirectUrl);
+
```
peschina, line 73: can we also check that the transaction status is correct in the payment request transaction list?

peschina (approved):
LGTM

apps/payments/e2e/specs/transactions/stipe.spec.ts
nnorbert: But then we will have to update it in every 20 years...  😄

peschina: Why worry 😂  Jokes apart, I don't mind keeping it as it is

nnorbert: I don't think that it is a big deal to calculate it dynamically, so I prefer this way honestly.

peschina (approved):

PR #860, opened 2024-08-22
Unit test messaging providers endpoints

Thurfjell:
### Ticket:

- [20683](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/20683)

### Description

Unit test erroruos endpoints sanitation for /api/v1/providers path.

Using mostly following pattern, making sure following scenarios fail
- field missing
- field empty (for string)
- field null type (and we assume this will work for the other valid json types, number, object, array, boolean, string)

There are also search query validation where appropriate as well as url parameter.

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [ ] **New feature**
- [x] **Dev change**
- [ ] **Additional tests**
- [ ] **Documentation**
- [ ] **Other**

### Checklist:

### Screenshots:

N/A

### DIFF
--- a/apps/messaging-api/src/app.ts
+++ b/apps/messaging-api/src/app.ts
@@ -28,17 +28,14 @@ const __dirname = dirname(__filename);
 dotenv.config();
 
 export async function build(opts?: FastifyServerOptions) {
-  const app = fastify(opts).withTypeProvider<TypeBoxTypeProvider>();
-  // app.setValidatorCompiler((schema) => {
-  //   const ajv = new Ajv({
-  //     removeAdditional: true,
-  //     useDefaults: true,
-  //     coerceTypes: false,
-  //     allErrors: true,
-  //   });
-
-  //   return ajv.compile.bind(ajv)(schema.schema);
-  // });
+  const app = fastify({
+    ...opts,
+    ajv: {
+      customOptions: {
+        coerceTypes: false,
+      },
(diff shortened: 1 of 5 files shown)


apps/messaging-api/src/app.ts
```diff
@@ -19,14 +19,24 @@ import fastifyMultipart from "@fastify/multipart";
...
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = dirname(__filename);
 
 dotenv.config();
 
 export async function build(opts?: FastifyServerOptions) {
-  const app = fastify(opts).withTypeProvider<TypeBoxTypeProvider>();
+  const app = fastify({
+    ...opts,
+    ajv: {
```
alfonsograziano, line 33: @peschina shall we apply the same in payments?

Thurfjell: Good that you raise it. @alfonsograziano 

I would recommend that we apply this rule on all APIs.

rgrnearform (approved):
lgtm

apps/messaging-api/src/app.ts
alfonsograziano: Agreed :)

peschina: yes!

PR #867, opened 2024-08-26
Messaging api change search query to string type with regex

Thurfjell:
### Ticket:

### Description

A url search query is always a string, and if we're not coercing values having a type guard for Number doesn't work since it's a string when it comes to schema validation. 

Thus the change is to use string with regex to match numeric. 

Note for the optional type, a "|" is necessary at the end of pattern to indicate that nothing is also ok.

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [ ] **New feature**
- [x] **Dev change**
- [ ] **Additional tests**
- [ ] **Documentation**
- [ ] **Other**

### Checklist:

### Screenshots:

N/A


edge33 (approved):
LGTM

PR #868, opened 2024-08-26
Feat/upload api file sharing

edge33:
### Ticket:

- [User Story 18796](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/18796): share file for view

### Description

- Added tables for sharing files, many files can be shared with many users

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**
- [ ] **Additional tests**
- [ ] **Documentation**
- [ ] **Other**

### Checklist:

### Screenshots:

N/A


apps/upload-api/db/migrations/sqls/20240822053657-file-sharing-table-up.sql
```diff
@@ -0,0 +1,19 @@
+BEGIN;
+
+-- Create the file_sharing table
+CREATE TABLE public.file_sharing (
+    id uuid NOT NULL, -- Foreign key to the files table
```
Thurfjell, line 5: Rename to file_id to avoid any confusion that this would just be a table id

apps/upload-api/db/migrations/sqls/20240822053657-file-sharing-table-up.sql
```diff
@@ -0,0 +1,19 @@
+BEGIN;
+
+-- Create the file_sharing table
+CREATE TABLE public.file_sharing (
```
Thurfjell, line 4: If this is the link table for files and users, rename to user_files.

apps/upload-api/src/routes/metadata/index.ts
```diff
@@ -0,0 +1,147 @@
...
+          "4xx": HttpError,
+          "5xx": HttpError,
+        },
+      },
+    },
+    async (request, reply) => {
+      const userId = ensureUserIdIsSet(request, METADATA_INDEX);
+      const organizationId = request.userData?.organizationId;
+
+      const data = await getUserFiles(app.pg, userId, organizationId);
```
Thurfjell, line 45: These don't include shared files right?

apps/upload-api/src/routes/metadata/index.ts
```diff
@@ -0,0 +1,147 @@
...
+      const userId = ensureUserIdIsSet(request, GET_METADATA);
+      const organizationId = request.userData?.organizationId;
+
+      const key = request.params.key;
+      const fileData = await getFileMetadata(
+        app.pg,
+        key,
+        userId,
+        organizationId,
+      );
```
Thurfjell, line 114: Both `getFileMetadata` and `getUserFiles` query the files table yes? Why two different files/methods?

apps/upload-api/src/routes/metadata/index.ts
```diff
@@ -0,0 +1,147 @@
...
+        ...f,
+        owner: usersData[f.ownerId],
+      }));
+
+      reply.send({ data: filesData });
+    },
+  );
+
+  app.get<{ Params: { key: string } }>(
+    "/:key",
```
Thurfjell, line 87: What does key mean? id?

apps/upload-api/src/routes/metadata/index.ts
edge33: not right now

apps/upload-api/src/routes/metadata/index.ts
edge33: the getFileMetadata gets the single file by id while the getUserFiles gets all the files a use has access to

Thurfjell: Can you rename functions so they match? getThing, getThings.

apps/upload-api/src/routes/utils/getUserFiles.ts
```diff
@@ -1,21 +1,67 @@
...
+    query = `${query} AND id NOT IN (${toExclude.map(() => `$${i++}`).join(", ")})`;
+  }
+
+  return connection.query<FileMetadataType>(query, [
+    organizationId,
+    ...toExclude,
+  ]);
+};
+
+export default async (
```
Thurfjell, line 39: I need to get my head around this.

Are you ever using this default export function without a org id?
Are these functions re-used in some permutation elsewhere? 
If you aren't using the org id you are "logged in with" (that's how it works right?) what org id are you ever going to pass into this function? Even if you use some other org id than your logged in one, you will get all your own files, together with this orgs file? Is this intentional behaviour?

apps/upload-api/src/routes/utils/getUserFiles.ts
```diff
@@ -1,21 +1,67 @@
...
+        organizationFiles = organizationFilesData.rows;
+      }
+    }
 
-  if (organizationId) {
-    query = `${query} OR organization_id = $2`;
-    params.push(organizationId);
+    return [...ownedFiles, ...organizationFiles];
+  } finally {
+    client.release();
```
Thurfjell, line 65: If this is meant to be reused, you could also have a client interface as param instead and handle the lifetime in the composing logic.

apps/upload-api/src/routes/utils/getUserFiles.ts
edge33: you can call it with:
- orgId if you are a public servant
- no orgId if you are a citizen

if the endpoint is accessed by a public servant, it will return all the files shared within the org and all the files the user owns, plus (in the future) all the files that are shared with the accessing user

Thurfjell: Can't you make separate requests instead of a catch all function? Now these queries aren't huge, however if we can get some cheap avoiding of the 3 months down the line "oh no we need X change for citizen files" situations. It's a bit more (duplicated?) code right, but if you can figure out which files to get depending on the request, call appropriate queries eg. isn't appropriate?

Note I'm happy to be convinced otherwise, just want to raise the discussion.

apps/upload-api/src/routes/metadata/index.ts
```diff
@@ -0,0 +1,170 @@
...
+          const organizationFilesData = await getOrganizationFiles(
+            client,
+            organizationId,
+            filesToExclude,
+          );
+
+          if (organizationFilesData.rows.length > 0) {
+            files.push(...organizationFilesData.rows);
+          }
+        }
```
Thurfjell, line 66: One last question. 

If you're an organisation user, you should never look for your own files right (should never any regardless)?

Thurfjell: That sounds crazy but I trust you are on top of business requirements.

Thurfjell (approved):

alfonsograziano (approved):

PR #881, opened 2024-08-28
feat(20776): create audit log events db table

nnorbert:
Create audit log events db table


apps/payments-api/utils/auditLogEvent.ts
```diff
@@ -0,0 +1,52 @@
...
+        audit_log_id as "auditLogId",
+        created_at as "createdAt",
+        event_type as "eventType",
+        user_id as "userId",
+        organization_id as "organizationId",
+        metadata
+    `,
+    [event.eventType, event.userId, event.organizationId, event.metadata],
+  );
+
```
nnorbert, line 50: Will have to check the result to not be empty.

alfonsograziano (dismissed):
LGTM - The idea you mentioned of creating a Fastify plugin so that we can inject it in the application is pretty good, feel free to do it here or in another PR

alfonsograziano (approved):
Great, thanks for the update - LGTM

PR #887, opened 2024-08-29
[Payments] Fix TrueLayer access token expiration

peschina:
### Ticket:

- [20872](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/20872)

### Description

Request to get TrueLayer access token was being cached, so we were getting stale tokens. We already make sure to call it only when the token has expired, so the request is now unique to make sure it's not cached
## Type

- [ ] **Dependency upgrade**
- [x] **Bug fix**
- [ ] **New feature**
- [ ] **Dev change**
- [ ] **Additional tests**
- [ ] **Documentation**
- [ ] **Other**

### Checklist:

### Screenshots:

N/A

### DIFF
--- a/apps/payments/app/integration/trueLayer.ts
+++ b/apps/payments/app/integration/trueLayer.ts
@@ -8,6 +8,7 @@ const CERTIFICATE_ID = process.env.TL_CERTIFICATE_ID ?? "";
 const PRIVATE_KEY = process.env.TL_PRIVATE_KEY ?? "";
 
 async function fetchNewAccessToken() {
+  // add a unique query param to avoid caching the fetch request
   const uniqueParam = new Date().getTime();
   const res = await fetch(
     `${process.env.TL_AUTH_SERVER_URI}/connect/token?unique=${uniqueParam}`,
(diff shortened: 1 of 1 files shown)


apps/payments/app/integration/trueLayer.ts
```diff
@@ -8,18 +8,22 @@ const CERTIFICATE_ID = process.env.TL_CERTIFICATE_ID ?? "";
 const PRIVATE_KEY = process.env.TL_PRIVATE_KEY ?? "";
 
 async function fetchNewAccessToken() {
-  const res = await fetch(`${process.env.TL_AUTH_SERVER_URI}/connect/token`, {
-    method: "POST",
-    headers: {
-      "Content-Type": "application/x-www-form-urlencoded",
+  const uniqueParam = new Date().getTime();
```
alfonsograziano, line 11: Makes sense to me, can we please add a comment explaining why we need the uniqueParam? (to avoid caching the fetch request)

peschina: [a6943f5](https://github.com/ogcio/life-events/pull/887/commits/a6943f54ce7540381e788c8d43381514567451cb)

nnorbert (approved):

alfonsograziano (approved):

PR #892, opened 2024-08-29
feat(20777): audit logs page

nnorbert:
audit logs page

### DIFF
--- a/apps/payments-api/plugins/auditLog/index.ts
+++ b/apps/payments-api/plugins/auditLog/index.ts
@@ -7,20 +7,9 @@ import fp from "fastify-plugin";
 import { AuditLogRepo } from "./repo";
 import { AuditLogEvent, AuditLogEventDO, CreateAuditLog } from "./types";
 import { PaginationParams } from "../../types/pagination";
+import { AuditLogEventTitles, AuditLogEventType } from "./auditLogEvents";
 
 export type AuditLogPlugin = Awaited<ReturnType<typeof buildPlugin>>;
-
-export enum AuditLogEventType {
-  TRANSACTION_CREATE = "transaction.create",
-  PROVIDER_CREATE = "provider.create",
-}
-
-export const AuditLogEventTitles: Record<string, string> & {
-  [key in AuditLogEventType]: string;
-} = {
-  [AuditLogEventType.TRANSACTION_CREATE]: "Transaction created",
-  [AuditLogEventType.PROVIDER_CREATE]: "Provider created",
-};
 
 const getEventTitle = (eventType: AuditLogEventType) => {
   return AuditLogEventTitles[eventType];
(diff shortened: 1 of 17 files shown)


apps/payments-api/plugins/auditLog/index.ts
```diff
@@ -5,13 +5,30 @@ import {
 } from "fastify";
 import fp from "fastify-plugin";
 import { AuditLogRepo } from "./repo";
-import { AuditLogEvent, CreateAuditLog } from "./types";
+import { AuditLogEvent, AuditLogEventDO, CreateAuditLog } from "./types";
+import { PaginationParams } from "../../types/pagination";
 
 export type AuditLogPlugin = Awaited<ReturnType<typeof buildPlugin>>;
 
+export enum AuditLogEventType {
```
alfonsograziano, line 13: I have the feeling that `AuditLogEventType` and `AuditLogEventTitles` might grow quite a lot in the future, shall we add those in a separate file to keep data and the logic separated?

apps/payments-api/routes/providers/index.ts
```diff
@@ -49,6 +50,15 @@ export default async function providers(app: FastifyInstance) {
         organizationId,
       );
 
+      app.auditLog.createEvent({
+        eventType: AuditLogEventType.PROVIDER_CREATE,
+        userId,
+        organizationId,
+        metadata: {
```
alfonsograziano, line 57: non-blocking, is great for now to have it as an example, we will need to decide what data to add in the audit log later and how to structure the relation with the resource (in this case the provider)

apps/payments/app/[locale]/(hosted)/paymentSetup/PaymentsMenu.tsx
```diff
@@ -24,7 +19,7 @@ export default ({ locale }: Props) => {
           <ol className={`govie-list govie-list--spaced ${styles.container}`}>
             <li tabIndex={0}>
               <Link
-                className="govie-button govie-button--icon govie-button--flat govie-button--icon govie-!-font-size-16"
+                className="govie-button govie-button--icon govie-button--flat govie-!-font-size-16"
```
Comment (alfonsograziano) [inline apps/payments/app/[locale]/(hosted)/paymentSetup/PaymentsMenu.tsx:22]: Are those changes related to this PR or needed? Just asking to be sure :)

alfonsograziano (dismissed):
LGTM - just left a couple of non-blocking comments

apps/payments-api/routes/providers/index.ts
nnorbert: This was temporary, just for testing purposes, I just forgot to remove it. Proper event creation will be implemented in a separate PR.

apps/payments/app/[locale]/(hosted)/paymentSetup/PaymentsMenu.tsx
Comment (nnorbert) [reply apps/payments/app/[locale]/(hosted)/paymentSetup/PaymentsMenu.tsx:22]: During my work I observed that this class appears twice on the links, which is nothing critical, but I thought it would be better to clean the code. It is too small for a separate PR, just a cleanup.

apps/payments-api/plugins/auditLog/index.ts
nnorbert: Done!

alfonsograziano (approved):

PR #894, opened 2024-08-30
feat(20782): filter audit logs

nnorbert:
Filter audit logs by event type.

### DIFF
--- a/apps/payments-api/openapi-definition.yml
+++ b/apps/payments-api/openapi-definition.yml
@@ -2552,7 +2552,7 @@ paths:
           description: Default Response
   /api/v1/auditLogs/:
     get:
-      tags:
+      tags: &a5
         - AuditLogs
       parameters:
         - schema:
@@ -2570,6 +2570,11 @@ paths:
             type: number
           in: query
           name: limit
+          required: false
+        - schema:
+            type: string
+          in: query
+          name: eventType
           required: false
       responses:
         "200":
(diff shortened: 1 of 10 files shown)


apps/payments-api/openapi-definition.yml
```diff
@@ -2552,7 +2552,7 @@ paths:
           description: Default Response
   /api/v1/auditLogs/:
     get:
-      tags:
+      tags: &a5
```
alfonsograziano, line 2555: This `&a5` looks weird, is it autogenerated?

apps/payments-api/plugins/auditLog/index.ts
```diff
@@ -43,12 +43,13 @@ const buildGetEvents =
   (repo: AuditLogRepo, log: FastifyBaseLogger) =>
   async (
     organizationId: string,
+    eventType: string | undefined,
```
alfonsograziano, line 46: Since we will need to add multiple filters, I suggest we create a new object to pass like "filters" which contains the eventType (and in the future the other filters)

apps/payments-api/openapi-definition.yml
nnorbert: Wow, I don't know what that is, but the file is autogenerated. I've never seen like this before. Is it some kind of reference? IDK, will check it.

nnorbert: Looks like it is a reference used to not repeat the tags array.

apps/payments-api/plugins/auditLog/index.ts
nnorbert: Done!

alfonsograziano (approved):
LGTM :)

apps/payments-api/openapi-definition.yml
alfonsograziano: Cool, good to know. Thanks!

PR #899, opened 2024-08-30
[Payments] audit log event detail page

peschina:
### Ticket:

- [20779](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/20779)

### Description

Added api and page to view the details of an audit log.

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**
- [ ] **Additional tests**
- [ ] **Documentation**
- [ ] **Other**

### Checklist:

### Screenshots:

N/A

### DIFF
--- a/apps/payments-api/plugins/auditLog/repo.ts
+++ b/apps/payments-api/plugins/auditLog/repo.ts
@@ -1,6 +1,7 @@
 import { PostgresDb } from "@fastify/postgres";
 import { AuditLogEvent, CreateAuditLog } from "./types";
 import { QueryResult } from "pg";
+import { PaginationParams } from "../../types/pagination";
 
 export class AuditLogRepo {
   pg: PostgresDb;
@@ -28,6 +29,41 @@ export class AuditLogRepo {
           metadata
       `,
       [event.eventType, event.userId, event.organizationId, event.metadata],
+    );
+  }
+
+  getEvents(
+    organizationId: string,
+    pagination: PaginationParams,
+  ): Promise<QueryResult<AuditLogEvent>> {
+    return this.pg.query(
+      `
+        SELECT
+          audit_log_id as "auditLogId",
+          created_at as "createdAt",
+          event_type as "eventType",
+          user_id as "userId",
+          organization_id as "organizationId"
+        FROM audit_logs
+        WHERE organization_id = $1
+        ORDER BY created_at DESC
+        LIMIT $2 OFFSET $3
+      `,
+      [organizationId, pagination.limit, pagination.offset],
+    );
+  }
+
+  getEventsTotalCount(
+    organizationId: string,
+  ): Promise<QueryResult<{ totalCount: number }>> {
+    return this.pg.query(
+      `
+        SELECT
+          count(*) as "totalCount"
+        FROM audit_logs
+        WHERE organization_id = $1
+      `,
+      [organizationId],
     );
   }
 }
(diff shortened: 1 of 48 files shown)


apps/payments-api/routes/auditLogs/index.ts
```diff
@@ -61,4 +70,38 @@ export default async function auditLogs(app: FastifyInstance) {
...
+          401: HttpError,
+          500: HttpError,
+        },
+      },
+    },
+    async (request, reply) => {
+      const organizationId = request.userData?.organizationId;
+      const { auditLogId } = request.params;
+
+      if (!organizationId) {
```
alfonsograziano, line 95: If `!organizationId`  I'd expect we throw an unauthorized error

peschina: good catch, I removed it by mistake [de53d54](https://github.com/ogcio/life-events/pull/899/commits/de53d54a1e2a4a319f7b0d413e3b80da8774d89c)

alfonsograziano (dismissed):
LGTM

apps/payments-api/plugins/auditLog/repo.ts
```diff
@@ -66,4 +66,21 @@ export class AuditLogRepo {
       [organizationId],
     );
   }
+
+  getEvent(eventId: string): Promise<QueryResult<AuditLogEventDetails>> {
```
nnorbert, line 70: I would say. to add the organizationID as well, because the user can see logs from its own organisation only. Sure, the ID is unique, but as an extra safety layer we should check orgID as well.

apps/payments-api/routes/auditLogs/index.ts
```diff
@@ -61,4 +70,39 @@ export default async function auditLogs(app: FastifyInstance) {
...
+    Params: ParamsWithAuditLogId;
+  }>(
+    "/:auditLogId",
+    {
+      preValidation: (req, res) =>
+        app.checkPermissions(req, res, [authPermissions.TRANSACTION_ALL]), // TODO: Change this later
+      schema: {
+        tags: TAGS_AUDIT_LOGS,
+        response: {
+          200: AuditLogEventDetails,
```
nnorbert, line 85: Please, to be consistent,, use the GenericResponse type.

apps/payments-api/routes/auditLogs/index.ts
```diff
@@ -61,4 +70,39 @@ export default async function auditLogs(app: FastifyInstance) {
...
+        throw app.httpErrors.unauthorized("Unauthorized!");
+      }
+
+      const eventDetails = await app.auditLog.getEventById(auditLogId);
+
+      if (eventDetails.organizationId !== organizationId) {
+        throw app.httpErrors.unauthorized("Unauthorized!");
+      }
+
+      reply.send(eventDetails);
```
nnorbert, line 105: Please send the data wrapped into a generic response.

apps/payments-api/plugins/auditLog/repo.ts
peschina: there's a check in the api, I can embed it in the query if you prefer

apps/payments-api/routes/auditLogs/index.ts
```diff
@@ -108,4 +110,39 @@ export default async function auditLogs(app: FastifyInstance) {
...
+      const organizationId = request.userData?.organizationId;
+      const { auditLogId } = request.params;
+
+      if (!organizationId) {
+        throw app.httpErrors.unauthorized("Unauthorized!");
+      }
+
+      const eventDetails = await app.auditLog.getEventById(auditLogId);
+
+      if (eventDetails.organizationId !== organizationId) {
```
nnorbert, line 141: This is not needed. We check if the currently logged in user has organisationID or not, but if the already saved event doesn't have any (which I think won't be possible) that doesn't mean that the user is unauthorised.

nnorbert (dismissed):
LGTM!

nnorbert (approved):

PR #908, opened 2024-09-02
Fix invalid provider total count

Thurfjell:
### Description

Exclude deleted in provider count

## Type

- [ ] **Dependency upgrade**
- [x] **Bug fix**
- [ ] **New feature**
- [ ] **Dev change**
- [ ] **Additional tests**
- [ ] **Documentation**
- [ ] **Other**

### Checklist:

### Screenshots:

N/A


apps/messaging-api/src/routes/providers/index.ts
SamSalvatico, line 1: You have to add all the where clauses as the "main" select
https://github.com/ogcio/life-events/blob/70cf726a98cf89abfabacefb652dc760ab1d55cc/apps/messaging-api/src/routes/providers/index.ts#L132

Thurfjell: It's already there what are you talking about. 

First clause of the cte is selecting count, the other is data.

SamSalvatico: The count query is not filtering for primary, or am I missing something?

Thurfjell: Should it?

SamSalvatico: Yes, it has to match exactly the query we do to get data, otherwise pagination links would be wrong

Thurfjell: updated

SamSalvatico (approved):

PR #911, opened 2024-09-03
[Payments] E2E tests - audit logs for provider created

peschina:
### Ticket:

- [21015](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/21015)

### Description

Added E2E test to make sure an audit log event is created when a new provider is created.
Changed metadata for provider creation to agreed structure. The same change will be implemented for other occurrences of audit log event creation.
Changed `manualBankTransfer` fixture to return name and id of the provider. The same logic will be applied to other resources created in fixtures if needed
## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [ ] **New feature**
- [x] **Dev change**
- [x] **Additional tests**
- [ ] **Documentation**
- [ ] **Other**

### Checklist:

### Screenshots:

N/A


apps/payments/e2e/utils/constants.ts
```diff
@@ -91,3 +91,35 @@ export const paymentRequestValidationErrorTexts: Record<
 };
 
 export const SAFE_PID_FILENAME = "ngrok.pid";
+
+export enum AuditLogEventType {
```
nnorbert, line 95: These were defined in the auditLogEvents file, which I can't see that it was deleted or not. Won't this code be duplicated?

peschina: They are in `payments-api,` not `payments` :)

rgrnearform (approved):

apps/payments/e2e/utils/constants.ts
nnorbert: I see. Thank you!

nnorbert (approved):
LGTM! The events duplication can remain for now, but we should find a better solution to not copy all the existing events from payments-api into payments. But that can be in a separate PR.

PR #920, opened 2024-09-04
21063 add filters to audit logs

nnorbert:
Add user filter and split event type into resource and action for filtering audit logs.


apps/profile-api/services/users/find-user.ts
```diff
@@ -71,25 +72,79 @@ export const findUser = async (params: {
...
 
-  const query = buildFindQuery([
-    "firstname ILIKE $1",
-    "lastname ILIKE $2",
-    "phone ILIKE $3",
-  ]);
+  let queryFields = ["phone ILIKE $1"];
+  let queryValues = [userParams.phone];
+
+  if (!userParams.strict) {
```
SamSalvatico, line 127:
```suggestion
  if (userParams.strict) {
```

SamSalvatico (changes requested):

apps/profile-api/services/users/find-user.ts
nnorbert: That means, if strict is set and true, then check the name. Is this what you want?

SamSalvatico (dismissed):

apps/payments-api/routes/auditLogs/index.ts
```diff
@@ -49,27 +49,60 @@ export default async function auditLogs(app: FastifyInstance) {
...
+        } else {
+          const [firstname, lastname] = user.split(" ");
+          findBy.firstname = firstname;
+          findBy.lastname = lastname;
+        }
+
+        const userDetails = await profileSdk.findUser(findBy);
+
+        if (userDetails.data === undefined) {
+          // ...
```
peschina, line 92: what here? :)

apps/payments-api/routes/auditLogs/index.ts
```diff
@@ -49,27 +49,60 @@ export default async function auditLogs(app: FastifyInstance) {
...
+        eventType: undefined,
+        userId: undefined,
+        from,
+        to,
+      };
+
+      filters.eventType = `${resource ?? "%"}.${action ?? "%"}`;
+
+      if (user) {
+        const profileSdk = await getProfileSdk(organizationId);
```
peschina, line 73: do you think it makes sense to move it to a service?

peschina (approved):
Looks good, a few comments

apps/payments-api/routes/auditLogs/index.ts
nnorbert: Most probably will be as we discussed, won't perform any search and return an empty result.

apps/payments-api/routes/auditLogs/index.ts
nnorbert: I can do that.

apps/payments-api/routes/auditLogs/index.ts
nnorbert: Done!

apps/payments-api/routes/auditLogs/index.ts
nnorbert: Done!

peschina (approved):

PR #926, opened 2024-09-04
Fix(messaging-api): 21103 - manage m2m sending messages

SamSalvatico:
### Ticket:

- [21103](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/21103)

### Description

Now also the M2M users can send messages

## Type

- [ ] **Dependency upgrade**
- [x] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**
- [ ] **Additional tests**
- [ ] **Documentation**
- [ ] **Other**


apps/messaging-api/src/routes/messages/index.ts
```diff
@@ -238,7 +238,12 @@ export default async function messages(app: FastifyInstance) {
         },
       },
     },
+
     async function getMessageHandler(request, _reply) {
+      console.log({
+        userIdHere: ensureUserIdIsSet(request, "TESTING"),
+        messageId: request.params.messageId,
+      });
```
Thurfjell, line 246: delete

packages/messages/migrations/sqls/20240904134403-add-sender-application-id-to-messages-up.sql
```diff
@@ -0,0 +1,12 @@
+BEGIN;
+
+ALTER TABLE messages
+ADD COLUMN sender_application_id varchar(21);
```
Thurfjell, line 4: I think this information should reside in the audit log and not in the message. I also think we should drop the sender information on message model. Not dying on that hill but this bothers e slightly.

apps/messaging-api/src/services/messages/messages.ts
```diff
@@ -615,6 +595,67 @@ export const processMessages = async (params: {
...
+    throw new BadRequestError(
+      errorProcess,
+      "You can't send messages to a different organization you are logged in to",
+    );
+  }
+  const toUseOrganizationId = organizationId ?? senderUser.organizationId;
+
+  if (isM2MApplicationSender) {
+    return { senderApplication: { id: senderUser.profileId } };
+  }
```
Thurfjell, line 633: Don't do this here, do it at composition level. 
Maybe rename getProfileUser() if you must have a function.
if(m2m) data = m2mdata
else data = profileData

Thurfjell (commented):
I think if the request identity is a m2m "admin" user, we could 
a) Create a new event type for creating message as m2m user
b) add relevant m2m fields to the create_message event type. 

Regardless, the only thing I'd like to see changed here, is of course being able to tell what identity the requester has (m2m or user), and set respective fields to existing logic.

packages/messages/migrations/sqls/20240904134403-add-sender-application-id-to-messages-up.sql
SamSalvatico: As a recipient I may need to know who sent this message, we already have the `sender_user_profile_id` info, then we need to add this here too

apps/messaging-api/src/services/messages/messages.ts
SamSalvatico: Sorry, don't get what you mean
To me it seems the method already does what you mean, renaming to `getProfileUser` would be wrong because application data are not related to user profile

SamSalvatico:
> I think if the request identity is a m2m "admin" user, we could a) Create a new event type for creating message as m2m user b) add relevant m2m fields to the create_message event type.
> 
> Regardless, the only thing I'd like to see changed here, is of course being able to tell what identity the requester has (m2m or user), and set respective fields to existing logic.

I updated the code to set the correct fields in the event log based on the sender type
Having a different event type would be logically wrong in my opinion because the message sending follows the same flow, then the `senderApplicationId`and the `senderUserProfileId` must be the fields we use to understand the sender type

apps/messaging-api/src/services/messages/messages.ts
Thurfjell: No I'm saying that you're creating a "let's solve multiple fundamentally different things in one abstraction". Create two abstraction if you need? If we keep creating abstraction, that keep solving multiple things at once, there's absolutely no gains to the abstraction to begin with except you have to function hop vs reading up to down in one place.

Thurfjell:
> > I think if the request identity is a m2m "admin" user, we could a) Create a new event type for creating message as m2m user b) add relevant m2m fields to the create_message event type.
> > Regardless, the only thing I'd like to see changed here, is of course being able to tell what identity the requester has (m2m or user), and set respective fields to existing logic.
> 
> I updated the code to set the correct fields in the event log based on the sender type Having a different event type would be logically wrong in my opinion because the message sending follows the same flow, then the `senderApplicationId`and the `senderUserProfileId` must be the fields we use to understand the sender type

Either have two event types, explicit for the purpose, one for created by m2m other by user, no need to have ambiguous data and you have to do null checks to figure out what created. Also not a huge hill but worth mentioning imo.

packages/messages/migrations/sqls/20240904134403-add-sender-application-id-to-messages-up.sql
Thurfjell: > I may need
Don't usually like this as argument for adding more fields to a table, however in this case I think it's something we care about? Possibly? I haven't ever heard that the citizen need to know which person sent the message. This isn't a p2p mail client.

apps/messaging-api/src/services/messages/messages.ts
SamSalvatico: Please let me know if this is what you meant
https://github.com/ogcio/life-events/pull/926/files#diff-c58b6982783a36a1f9cdad1d64a1d5fa3f0f12272e28b77c871ba7cb98ec2a6cR537

packages/messages/migrations/sqls/20240904134403-add-sender-application-id-to-messages-up.sql
SamSalvatico: I think possible benefits are higher than cons in this case

SamSalvatico: @Thurfjell what do you think about maintaining this for this PR and then remove it in a next pr also removing the sender user profile I'd if not needed?

Thurfjell (approved):
Let's go

PR #928, opened 2024-09-05
create message happy e2e api test

Thurfjell:
### Ticket:

- [21199](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/21199)

### Description

Create message e2e api test happy path via imported user

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [ ] **New feature**
- [x] **Dev change**
- [ ] **Additional tests**
- [ ] **Documentation**
- [ ] **Other**

### Checklist:

### Screenshots:

N/A


SamSalvatico (approved):

apps/messaging-api/e2e/PublicServant/Messages/Import user.bru
```diff
@@ -0,0 +1,50 @@
...
+    "collectedConsent":true,
+    "lastName":"Testsson",
+    "firstName":"Test"
+  }]
+}
+
+script:post-response {
+  const axios = require('axios')
+  
+  const importRequest = await axios(`${bru.getEnvVar("baseUrl")}/api/v1/user-imports`,{
```
edge33, line 30: is this an axios get to the same endpoint to retrieve the imported data?

apps/messaging-api/e2e/PublicServant/Messages/Find created message event.bru
```diff
@@ -0,0 +1,23 @@
...
+  search: {{messageSubject}}
+}
+
+auth:bearer {
+  token: {{accessToken}}
+}
+
+script:post-response {
+  bru.setVar("eventId",res.body.data[0].id)
+}
```
edge33, line 23: no assertion?

edge33 (dismissed):
LGTM, the .bru files are missing assertions but if that's expected I'm ok with it

apps/messaging-api/e2e/PublicServant/Messages/Import user.bru
SamSalvatico: @Thurfjell I would not do this, instead setting a collection variable getting the returned user import id, let's say {{userImportId}} and then, in a following test file, use that var to get the imported import

SamSalvatico: As we do for Public Servant/Templates

SamSalvatico (changes requested):

apps/messaging-api/e2e/PublicServant/Messages/Import user.bru
edge33: so the api should return the imported user ids?

SamSalvatico: I checked and it doesn't return anything
Anyway, I would do the requests not with axios but as following tests
You can do the same tests but using other Bru files, in this way the flow will be clearer

Thurfjell: So I was contemplating wether we should have 10 "tests" (setup) per suite, or just pre/post scripting them.

If we're not seeding, we need to setup with api calls. 

The goal of this suite would be to verify that the message gets created with expected content and other fields. Not wether import user, get user by import id eg. works. 

If we want a more holistic approach, I would like us to jointly solve that.

apps/messaging-api/e2e/PublicServant/Messages/Find created message event.bru
Thurfjell: Not testing the endpoint. 

Should we? 

This goes to above discussion, should we create tests (in a suite) for each call we have to make, but omit any assertions, or should we do them in pre/post scripts, just assuming they work since they have suites covering their functionality elsewhere. 

I'm not as sold on this is the wrong way as @SamSalvatico is.

Thurfjell:
> As we do for Public Servant/Templates

What does this mean @SamSalvatico

apps/messaging-api/e2e/PublicServant/Messages/Import user.bru
Thurfjell: I'd go as far to the other direction, that we'd do all the setup in pre/post script, in the "create message" test.

SamSalvatico: I would avoid doing Axios requests to invoke other APIs related to the current one
We can use collection variables to test needed values. Let's say that you want to import a user with email `a@b.com`, you can set a pre-request var, use it in the body and then, in a following request, check if a user with that email address exists

As you suggested we can create folders for each specific use case we need to test, but in both Pre & Post requests scripts I would only test data related to the current request itself

Thurfjell: I don't want to verify anything except that my message I posted, I somehow guaranteed to be represented in some persistence. I don't care about having a bunch of "did user import work" or a trail of importing users, when I create a message.

apps/messaging-api/e2e/PublicServant/Messages/Find created message event.bru
SamSalvatico: I would do assertions related to the specific endpoint
At least some minimum, like 200 status code and response.body.data.length > 0
In this way we will break the circuit just when something is not working, having clearer error messages and avoiding doing following requests

apps/messaging-api/e2e/PublicServant/Messages/Import user.bru
Thurfjell: > I would avoid doing Axios requests to invoke other APIs related to the current one

Why though, it's part of a pre-request setup script.

SamSalvatico:
> > As we do for Public Servant/Templates
> 
> What does this mean @SamSalvatico


I wrote the comment in the wrong thread, sorry
Anyway, when we get a list of templates through E2E we set a variable with the first template id got and we invoke the GET template one with that id

apps/messaging-api/e2e/PublicServant/Messages/Import user.bru
SamSalvatico: Then we should update our API to return a userImportId instead of null, so we can test the got body response and do other tests in other files

If you agree I will try setting the response body with id

Thurfjell:
I'll restructure this with some further folder structure, and we keep each http setup call as a test with some rudimentary "it should only have the id we're looking for", not to necessarily assert response, but to circuit break so it's clearer where. 

Sounds good @SamSalvatico  @edge33 ?

SamSalvatico:
> I'll restructure this with some further folder structure, and we keep each http setup call as a test with some rudimentary "it should only have the id we're looking for", not to necessarily assert response, but to circuit break so it's clearer where.
> 
> Sounds good @SamSalvatico @edge33 ?

It totally does for me

SamSalvatico (approved):
Now I definitely like it

Thank you @Thurfjell

PR #957, opened 2024-09-11
Chore(messaging): updated preferred language management

SamSalvatico:
### Description

- Fixed the permissions management on the messaging app to get the correct profile token
- get the preferred language from select API
- use the preferred language to send invitations

## Type

- [ ] **Dependency upgrade**
- [x] **Bug fix**
- [x] **New feature**
- [x] **Dev change**

### DIFF
--- a/apps/messaging-api/src/services/users/invitations/send-invitations.ts
+++ b/apps/messaging-api/src/services/users/invitations/send-invitations.ts
@@ -14,6 +14,7 @@ import { ALL_TRANSPORTS } from "../shared-users.js";
 import { ServerError } from "shared-errors";
 import { CreateMessageParams } from "../../messages/messaging.js";
 import { processMessages } from "../../messages/messages.js";
+import { getProfileSdk } from "../../../utils/authentication-factory.js";
 
 const SEND_INVITATIONS_ERROR = "SEND_INVITATIONS_ERROR";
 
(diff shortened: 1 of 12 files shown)


apps/messaging-api/src/services/users/invitations/send-invitations.ts
```diff
@@ -335,22 +337,36 @@ const sendInvitations = async (params: {
...
-const getLanguagePerUser = (params: {
+const getLanguagePerUser = async (params: {
   userIdsToSearchFor: { userProfileId: string; userId: string }[];
   requestUserId: string;
-}): { [x: string]: string } => {
+  organisationId: string;
+}): Promise<{ [x: string]: string }> => {
   if (params.userIdsToSearchFor.length === 0) {
     return {};
   }
```
Thurfjell, line 347: I wonder if we should  throw here? This feels like an illegal parameter. Especially if it's chain called via spread operators `getstuff({ ...GetLanguagePerUser(...otherstuff)})`

wdyt?

apps/messaging-api/src/services/users/invitations/send-invitations.ts
```diff
@@ -335,22 +337,36 @@ const sendInvitations = async (params: {
...
+
+  if (profiles.error || !profiles.data) {
+    return output;
+  }
+
+  for (const userProfile of profiles.data) {
+    const userId = mappedIds[userProfile.id];
+    output[userId] = userProfile.preferredLanguage;
+  }
```
Thurfjell, line 368: 🙏

Thurfjell (approved):

Thurfjell (dismissed):
🌟

Thurfjell (approved):

PR #965, opened 2024-09-13
Messaging API - Message actions route seen/unseen

Thurfjell:
### Ticket:

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**
- [ ] **Additional tests**
- [ ] **Documentation**
- [ ] **Other**

### Checklist:

### Screenshots:

N/A


apps/messaging-api/src/routes/message-actions/index.ts
```diff
@@ -0,0 +1,144 @@
...
+            isSeen: boolean;
+          }>
+        | undefined;
+
+      const jobEvents: { type: EventType; data: MessageEventData }[] = [];
+      let poolClient: PoolClient | undefined;
+      try {
+        poolClient = await app.pg.pool.connect();
+        console.log("broder tuck");
+        console.log(userId, req.body.messageId);
```
SamSalvatico, line 76:
```suggestion
```

apps/messaging-api/src/routes/message-actions/index.ts
```diff
@@ -0,0 +1,144 @@
...
+        updateQueryResult = await poolClient.query<{ isSeen: boolean }>(
+          `
+            with message_selection as(
+                select m.id from messages m
+                join users u on (u.user_profile_id = $1 or u.id::text = $1)
+                where 
+                    (u.id::text = $1 or u.user_profile_id = $1)
+                    and is_delivered = true 
+                    and m.id = $2
+                    and $3 != m.is_seen
```
SamSalvatico, line 105: Do we want to return an error if an already seen message is set as seen again?

SamSalvatico (commented):
Few comments, but great work

apps/messaging-api/src/routes/message-actions/index.ts
Thurfjell: No with motivation:

(big) if we add more actions, we'll treat this as an ordinary put. There's nothing inherently criminal with putting a resource to the same. Just nothing happens. I think this is normal behaviour.

SamSalvatico: Sorry, I misread the code totally fine

SamSalvatico (approved):

PR #971, opened 2024-09-16
Chore(messaging-api): 21497 add error event log

SamSalvatico:
### Ticket:

- [21497](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/21497)

### Description

Added error details on event log
## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [ ] **New feature**
- [x] **Dev change**
- [ ] **Additional tests**
- [ ] **Documentation**
- [ ] **Other**

### DIFF
--- a/apps/messaging-api/src/routes/providers/services.ts
+++ b/apps/messaging-api/src/routes/providers/services.ts
@@ -1,6 +1,5 @@
 import nodemailer from "nodemailer";
 import { PoolClient } from "pg";
-import { ServiceError } from "../../utils.js";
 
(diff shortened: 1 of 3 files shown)


Thurfjell (approved):
We would probably want to refine what we show, rather than the err dump, if possible, in the future? This is great though.

SamSalvatico:
> We would probably want to refine what we show, rather than the err dump, if possible, in the future? This is great though.

I agree with you, but at the moment is not shown on UI

PR #980, opened 2024-09-18
feat(scheduler-api) Added config table and scheduler callback endpoint

edge33:
### Ticket:

- [User Story 18797](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/18797): delete files

### Description

- Added config table to store settings
- Added callback url for scheduler jobs

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**
- [ ] **Additional tests**
- [ ] **Documentation**
- [ ] **Other**

### Checklist:

### Screenshots:

N/A

### DIFF
--- a/apps/upload-api/db/migrations/sqls/20240918075102-add-settings-table-up.sql
+++ b/apps/upload-api/db/migrations/sqls/20240918075102-add-settings-table-up.sql
@@ -1,0 +1,9 @@
+CREATE TABLE settings (
+    id SERIAL PRIMARY KEY,
+    key VARCHAR(255) NOT NULL UNIQUE,
+    value TEXT NOT NULL,
+    type VARCHAR(50),  -- Optional: Can specify the type like string, boolean, integer, etc.
+    description TEXT,  -- Optional: Used to explain the setting
+    created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    updated_at TIMESTAMPTZ NOT NULL DEFAULT NOW()
+);
--- a/apps/upload-api/db/migrations/sqls/20240918075102-add-settings-table-up.sql
+++ b/apps/upload-api/db/migrations/sqls/20240918075102-add-settings-table-up.sql
@@ -2,8 +2,8 @@ CREATE TABLE settings (
     id SERIAL PRIMARY KEY,
     key VARCHAR(255) NOT NULL UNIQUE,
     value TEXT NOT NULL,
-    type VARCHAR(50),  -- Optional: Can specify the type like string, boolean, integer, etc.
-    description TEXT,  -- Optional: Used to explain the setting
+    type TEXT,
+    description TEXT,
     created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
     updated_at TIMESTAMPTZ NOT NULL DEFAULT NOW()
 );
--- a/apps/upload-api/src/routes/scheduler.ts
+++ b/apps/upload-api/src/routes/scheduler.ts
@@ -2,18 +2,18 @@ import { FastifyInstance } from "fastify";
 import { getConfigValue } from "../utils/storeConfig.js";
 
 export default async function schduler(app: FastifyInstance) {
-  app.post<{ Body: { callbackUuid: string } }>(
+  app.post<{ Body: { schedulerToken: string } }>(
     "/",
     { schema: { hide: true } },
     async (request) => {
-      const { callbackUuid } = request.body;
+      const { schedulerToken } = request.body;
 
-      const expectedUuid = await getConfigValue(
+      const expectedSchedulerToken = await getConfigValue(
         app.pg.pool,
-        "schedulerCallback",
+        "schedulerToken",
       );
 
-      if (callbackUuid !== expectedUuid) {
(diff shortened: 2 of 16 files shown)


apps/upload-api/db/migrations/sqls/20240918075102-add-settings-table-up.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE settings (
+    id SERIAL PRIMARY KEY,
+    key VARCHAR(255) NOT NULL UNIQUE,
```
Thurfjell, line 3: Why are we using varchar eg. ? Text is iirc the preferred way in almost all cases when using Postgres?

apps/upload-api/db/migrations/sqls/20240918075102-add-settings-table-up.sql
```diff
@@ -0,0 +1,9 @@
...
+    value TEXT NOT NULL,
+    type VARCHAR(50),  -- Optional: Can specify the type like string, boolean, integer, etc.
+    description TEXT,  -- Optional: Used to explain the setting
+    created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    updated_at TIMESTAMPTZ NOT NULL DEFAULT NOW()
```
Thurfjell, line 8: Super duper nit, but have you updated something if it wasn't already created?

apps/upload-api/src/routes/scheduler.ts
```diff
@@ -0,0 +1,25 @@
+import { FastifyInstance } from "fastify";
+import { getConfigValue } from "../utils/storeConfig.js";
+
+export default async function schduler(app: FastifyInstance) {
+  app.post<{ Body: { callbackUuid: string } }>(
```
Thurfjell, line 5: If this is the token used for identification, maybe callbackToken to be clear about intentions?

Thurfjell: Or even key as you use in your database column

apps/upload-api/src/test/utils/storeConfig.test.ts
```diff
@@ -0,0 +1,85 @@
+import t from "tap";
+import { Pool } from "pg";
+import {
+  CONFIG_TYPE,
+  getConfigValue,
+  storeConfig,
+} from "../../utils/storeConfig.js";
+
+t.test("storeConfig", async (t) => {
```
Thurfjell, line 9: What's the purpose of these unit tests? It's just a query. I'd rather see this as e2e/integration to verify the validity of your query if that's necessary.

apps/upload-api/db/migrations/sqls/20240918075102-add-settings-table-up.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE settings (
```
Thurfjell, line 1: What kind of settings is this? `settings.type`, is very abstract to me.

apps/upload-api/src/test/utils/storeConfig.test.ts
edge33: totally, right, I added the test for the coverage, I will add e2e for the actual use case when the endpoint is ready

apps/upload-api/db/migrations/sqls/20240918075102-add-settings-table-up.sql
edge33: application settings, type is the type of value as it should be parsed when retrieved from the db

apps/upload-api/src/routes/scheduler.ts
```diff
@@ -0,0 +1,25 @@
...
+  app.post<{ Body: { schedulerToken: string } }>(
+    "/",
+    { schema: { hide: true } },
+    async (request) => {
+      const { schedulerToken } = request.body;
+
+      const expectedSchedulerToken = await getConfigValue(
+        app.pg.pool,
+        "schedulerToken",
+      );
```
Thurfjell, line 14: Correct me if I'm in the wrongs here. 

I think the idea of having a token for identifying a callback request, Is that you persist one per "job" right. Perform some scheduled action on file.id = A, you would create a token for A, and possibly describe some sort of action (delete, archive.. whatever), and maybe not store a token for all operations, right?

apps/upload-api/src/test/utils/storeConfig.test.ts
Thurfjell: Why do we need test for coverage, If we're not convinced they actually help? We have a metric quota?

apps/upload-api/src/routes/scheduler.ts
edge33: it might be a good approach, but for the upload-api we only want a single job that deletes file scheduled for deletion, so it will only have one token for now

apps/upload-api/src/test/utils/storeConfig.test.ts
edge33: not really, but I rather have the pipeline silent 🤫

Thurfjell (approved):

PR #1017, opened 2024-10-01
working implementation of communication RFC

alfonsograziano:
### Ticket:

- [22072](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/22072)

### Description

Implemented secure communication RFC 

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [X] **New feature**
- [ ] **Dev change**
- [ ] **Additional tests**
- [ ] **Documentation**
- [ ] **Other**

### Checklist:

### Screenshots:

N/A

### DIFF
--- a/apps/integrator-api/app.ts
+++ b/apps/integrator-api/app.ts
@@ -13,7 +13,7 @@ import { fileURLToPath } from "url";
 import { dirname, join } from "path";
 import sensible from "@fastify/sensible";
 import schemaValidators from "./src/routes/schemas/validations";
-import apiAuthPlugin from "api-auth";
+import apiAuthPlugin, { verifyJWT } from "api-auth";
 import { initializeErrorHandler } from "error-handler";
 import { initializeLoggingHooks } from "logging-wrapper";
 import healthCheck from "./src/routes/healthcheck";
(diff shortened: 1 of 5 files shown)


apps/integrator-api/app.ts
```diff
@@ -84,6 +84,38 @@ export async function build(opts?: FastifyServerOptions) {
...
+    const { token } = request.query as any;
+
+    if (!token) {
+      return reply.code(400).send({ error: "Token not provided" });
+    }
+
+    // We need to store this somehow in an env variable
+    const jwksRegistry = {
+      payments: "http://localhost:8001/.well-known/jwks.json",
+    };
```
peschina, line 99: We would need just the payments domain (or similar) in an env variable

apps/payments-api/app.ts
```diff
@@ -98,5 +102,42 @@ export async function build(opts?: FastifyServerOptions) {
...
+    const { publicKey } = await readOrGenerateKeyPair("payments-api");
+    return getJWKSRoute(publicKey);
+  });
+
+  // This is a test route that calls the integrator with a callback
+  app.get("/call-integrator", async () => {
+    // const privateKey = await readLocalPrivateKey(
+    //   join(__dirname, "private.key"),
+    // );
+    const { privateKey } = await readOrGenerateKeyPair("payments-api");
```
peschina, line 117: How will we actually get the private key?

peschina (approved):
Looks good, but we need a suitable solution for key storage and rotation

apps/integrator-api/app.ts
alfonsograziano: Yess

apps/payments-api/app.ts
alfonsograziano: This method is meant to be used only locally (I don't want people to create multiple keypair on their machine). In the new PR to integrate it properly within payments, I'll manage that with KMS with @williamwd

PR #1022, opened 2024-10-01
[22130] Integrator APIs for journey

peschina:
### Ticket:

- [22130](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/22130)

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [ ] **New feature**
- [x] **Dev change**
- [ ] **Additional tests**
- [ ] **Documentation**
- [ ] **Other**

### Checklist:

### Screenshots:

N/A


apps/integrator-api/src/plugins/entities/journey/repo.ts
```diff
@@ -1,6 +1,15 @@
 import { PostgresDb } from "@fastify/postgres";
 import { QueryResult } from "pg";
-import { JourneyDetailsDO } from "../../../routes/schemas";
+import {
+  CreateJourneyBodyDO,
+  JourneyDetailsDO,
+  JourneyStatusType,
+} from "../../../routes/schemas";
+
+enum JourneyStatus {
```
nnorbert, line 9: Let's make this enum global and shared.

apps/integrator-api/src/routes/journeys/index.ts
```diff
@@ -23,17 +62,127 @@ export default async function journeys(app: FastifyInstance) {
...
+        throw app.httpErrors.unauthorized("Unauthorized!");
+      }
+
+      const res = await app.journey.createJourney(request.body);
+
+      reply.send(formatAPIResponse(res));
+    },
+  );
+
+  app.patch<{
```
nnorbert, line 151: Just to be consistent, we should use the same method. PATCH is the correct one if we take the method's description into consideration, but in the API description, there is a PUT. Somehow, that also makes sense since we are allowing all the fields to be edited, which makes sense to change (id, created_at, etc. can't be changed), and we are requesting all of those fields, even though they weren't changed. So... for me, it is the same, to be honest.

apps/integrator-api/src/routes/schemas/index.ts
```diff
@@ -37,15 +39,96 @@ export type GenericResponse<T> = {
...
+export type JourneyStep = Static<typeof JourneyStep>;
+
+export const JourneyStepConnection = Type.Object({
+  id: Type.String(),
+  sourceStepId: Type.String(),
+  destinationStepId: Type.String(),
+});
+export type JourneyStepConnection = Static<typeof JourneyStepConnection>;
+
+export const JourneyDetails = Type.Composite([
```
nnorbert, line 103: This is just my opinion, but I would do the other way around:
- I have the JourneyDetails with all the fields
- I have a JourneyPublicDetails, which is a subset (use Type.Pick) of the JourneyDetails

I'm thinking like, to have the base type, that in our case has everything, and I need something more limited, so I am picking the fields that are needed for my subset.

nnorbert (approved):
I think, it looks good, only some comments with my personal points of view.

apps/integrator-api/src/routes/journeys/index.ts
peschina: I would keep PATCH and allow only editing of the fields that are intended to be updated for this endpoint. In the future we can change to PUT or add it if the use case arises. Wdyt?

nnorbert: It is okay, but in a patch every field should be optional. That's the main difference between the two methods. I think, the declared fields are required in the body.

peschina: you have a point, let's go with PUT!

peschina: [1c1747f](https://github.com/ogcio/life-events/pull/1022/commits/1c1747f9de02915e88b368f779ad880e1c0ea9d0)

PR #1024, opened 2024-10-02
22130 integrator api steps linkage

nnorbert:
Journey step connections.

### DIFF
--- a/apps/integrator/app/types/index.ts
+++ b/apps/integrator/app/types/index.ts
@@ -1,12 +1,5 @@
 import { PluginData } from "./plugins";
 
-export type Journey = {
-  id: string;
-  title: string;
-  createdAt: Date;
-  updatedAt: Date;
-};
-
 export enum STEP_TYPE {
   FORM = "form",
   PAYMENT = "payment",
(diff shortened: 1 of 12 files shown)


apps/integrator/app/types/index.ts
```diff
@@ -47,3 +40,22 @@ export type Submission = {
...
+  destinationStepId: string;
+};
+
+export type Journey = {
+  id: string;
+  title: string;
+  userId: string;
+  organizationId: string;
+  status: string;
+  initialStepId: string;
```
peschina, line 56: shouldn't `initialStepId` be optional?

peschina (approved):
Looks good, just one comment

apps/integrator/app/types/index.ts
nnorbert: Yes, it should, good catch! Thank you!

nnorbert: Done

peschina (approved):

alfonsograziano (approved):
LGTM!

PR #1033, opened 2024-10-07
[Integrator] Executor - GET APIs

peschina:
### Ticket:

- [22309](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/22309)

### Description

Added GET APIs for executor

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**
- [ ] **Additional tests**
- [ ] **Documentation**
- [ ] **Other**

### Checklist:

### Screenshots:

N/A

### DIFF
--- a/apps/integrator-api/src/plugins/entities/run/repo.ts
+++ b/apps/integrator-api/src/plugins/entities/run/repo.ts
@@ -1,0 +1,27 @@
+import { PostgresDb } from "@fastify/postgres";
+import { QueryResult } from "pg";
+import { RunDO } from "./types";
+
+export class RunRepo {
+  pg: PostgresDb;
+
+  constructor(pg: PostgresDb) {
+    this.pg = pg;
+  }
+
+  getRuns(userId: string): Promise<QueryResult<RunDO>> {
+    return this.pg.query(
+      `SELECT
+        id,
+        status,
+        journey_id as "journeyId",
+        created_at as "createdAt",
+        updated_at as "updatedAt",
+        user_id as "userId"
+      FROM runs
+      WHERE user_id = $1
+      ORDER BY created_at DESC`,
+      [userId],
+    );
+  }
+}
--- a/apps/integrator-api/src/plugins/entities/run/repo.ts
+++ b/apps/integrator-api/src/plugins/entities/run/repo.ts
@@ -1,6 +1,6 @@
 import { PostgresDb } from "@fastify/postgres";
 import { QueryResult } from "pg";
-import { RunDetailsDO } from "./types";
+import { RunDetailsDO, RunStepDO } from "./types";
 
 export class RunRepo {
   pg: PostgresDb;
@@ -25,7 +25,10 @@ export class RunRepo {
     );
   }
 
-  getRunsByJourney(journeyId: string): Promise<QueryResult<RunDetailsDO>> {
+  getRunsByJourney(
+    journeyId: string,
+    organizationId: string,
+  ): Promise<QueryResult<RunDetailsDO>> {
     return this.pg.query(
       `SELECT
         id,
@@ -35,9 +38,9 @@ export class RunRepo {
         updated_at as "updatedAt",
         user_id as "userId"
       FROM runs
-      WHERE journey_id = $1
+      WHERE journey_id = $1 and organization_id= $2
       ORDER BY created_at DESC`,
-      [journeyId],
+      [journeyId, organizationId],
     );
   }
 
@@ -60,7 +63,10 @@ export class RunRepo {
     );
   }
 
-  getRunById(runId: string): Promise<QueryResult<RunDetailsDO>> {
+  getRunById(
+    runId: string,
+    organizationId: string,
+  ): Promise<QueryResult<RunDetailsDO>> {
     return this.pg.query(
       `SELECT
               id,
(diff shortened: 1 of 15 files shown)


apps/integrator-api/src/plugins/entities/run/repo.ts
```diff
@@ -0,0 +1,94 @@
...
+              status,
+              created_at as "createdAt",
+              updated_at as "updatedAt",
+            FROM runs
+            WHERE id = $1`,
+      [runId],
+    );
+  }
+
+  getRunStepsByRunId(runId: string): Promise<QueryResult<any>> {
```
alfonsograziano, line 79: Is it correct to have an `any` here?

apps/integrator-api/src/routes/executor/index.ts
```diff
@@ -0,0 +1,151 @@
...
+    },
+    async (request, reply) => {
+      const { journeyId } = request.params;
+      const organizationId = request.userData?.organizationId;
+
+      if (!organizationId) {
+        throw app.httpErrors.unauthorized("Unauthorized!");
+      }
+
+      const runs = await app.run.getRunsByJourneyId(journeyId);
```
alfonsograziano, line 108: From what I can read, it looks like everyone from every organization can have access to all the runs just by providing a valid journey id, most probably we want to add a second `organizationId` param to enforce that we're filtering for the current org.

apps/integrator-api/src/routes/executor/index.ts
```diff
@@ -0,0 +1,151 @@
...
+    },
+    async (request, reply) => {
+      const { runId } = request.params;
+      const organizationId = request.userData?.organizationId;
+
+      if (!organizationId) {
+        throw app.httpErrors.unauthorized("Unauthorized!");
+      }
+
+      const runDetails = await app.run.getRunById(runId);
```
alfonsograziano, line 139: Same here, most probably we want to be able to read all the runs connected to a journey that belongs to our current organization

apps/integrator-api/src/routes/executor/index.ts
```diff
@@ -0,0 +1,151 @@
...
+
+      const steps = await app.run.getRunStepsByRunId(runId);
```
alfonsograziano, line 141: Non-blocking: we should be able to parallelize the two queries

peschina: [a06a4a6](https://github.com/ogcio/life-events/pull/1033/commits/a06a4a638b06f89582e18ab1bb985279423ad804)

apps/integrator-api/src/routes/executor/index.ts
peschina: [a06a4a6](https://github.com/ogcio/life-events/pull/1033/commits/a06a4a638b06f89582e18ab1bb985279423ad804)

apps/integrator-api/src/plugins/entities/run/repo.ts
peschina: good catch [a06a4a6](https://github.com/ogcio/life-events/pull/1033/commits/a06a4a638b06f89582e18ab1bb985279423ad804)

apps/integrator-api/src/routes/executor/index.ts
```diff
@@ -0,0 +1,153 @@
...
+      const { runId } = request.params;
+      const userId = request.userData?.userId;
+
+      if (!userId) {
+        throw app.httpErrors.unauthorized("Unauthorized!");
+      }
+
+      const [runDetails, steps] = await Promise.all([
+        app.run.getUserRunById(runId, userId),
+        app.run.getRunStepsByRunId(runId),
```
alfonsograziano, line 73: Do we have to add the `userId` here as well?

apps/integrator-api/src/routes/executor/index.ts
```diff
@@ -0,0 +1,153 @@
...
+      const { runId } = request.params;
+      const organizationId = request.userData?.organizationId;
+
+      if (!organizationId) {
+        throw app.httpErrors.unauthorized("Unauthorized!");
+      }
+
+      const [runDetails, steps] = await Promise.all([
+        app.run.getRunById(runId, organizationId),
+        app.run.getRunStepsByRunId(runId),
```
alfonsograziano, line 142: Same here

alfonsograziano (approved):
LGTM - just left 2 comments, I'm not 100% sure, but I think it might make sense

apps/integrator-api/src/routes/executor/index.ts
peschina: We don't store `userId` neither `organizationId` in the `run_steps` table. Do you think it would be really needed? The methods is used in these apis only so far, and the API would return an error anyways if you don't have permission to read that run

apps/integrator-api/src/routes/executor/index.ts
peschina: Same as above

apps/integrator-api/src/routes/executor/index.ts
alfonsograziano: In order to check the userId, we can join the steps with the run and from there check the user id. If we're sure that the API returns an error if you don't have the right level of permission to read it, then I'm fine with it

nnorbert (approved):
LGTM

PR #1037, opened 2024-10-07
Payments token

alfonsograziano:
### Ticket:

- [22311](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/22311)

### Description

Appending signed token when calling the redirectUrl

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [X] **New feature**
- [ ] **Dev change**
- [ ] **Additional tests**
- [ ] **Documentation**
- [ ] **Other**


packages/api-auth/jwtService.ts
```diff
@@ -85,6 +85,8 @@ async function createSignedJWT(
     issuer,
   } = options;
 
+  const pk = await importJWK(privateKey, algorithm);
```
alfonsograziano, line 88: Copy of @peschina PR

apps/payments-api/routes/redirectToken/index.ts
peschina, line 1: Could this endpoint be under the transactions route?

apps/payments-api/routes/schemas/index.ts
```diff
@@ -476,3 +476,7 @@ export const AuditLogEventsFiltersQueryString = Type.Composite([
 export type AuditLogEventsFiltersQueryString = Static<
   typeof AuditLogEventsFiltersQueryString
 >;
+
+export const RedirectTokenObject = Type.Object({
```
peschina, line 480: Shall we name this more generically `TokenObject` so we can reuse the type?

peschina (commented):
Looks good, a few comments

apps/payments-api/routes/redirectToken/index.ts
```diff
@@ -0,0 +1,59 @@
...
+    },
+    async (request, reply) => {
+      const userId = request.userData?.userId;
+      const { transactionId } = request.params;
+
+      if (!userId) {
+        throw app.httpErrors.unauthorized("Unauthorized!");
+      }
+
+      const transactionDetails = await app.transactions.getTransactionById(
```
nnorbert, line 39: Why do we need to load the transaction details, when all the data we are using is the transactionId, which is already comes in the parameters.

nnorbert: Just to validate, that transaction exists? Then it is okay. We can optimise it to write a specific query for this action.

nnorbert (approved):
The code looks good, but the flow is not clear. We can discuss it in private.

apps/payments-api/routes/redirectToken/index.ts
alfonsograziano: That's correct :) We want to validate that the transaction exists. Also, for now, we're passing only the transactionId, but we might need to pass other data as well

apps/payments-api/routes/schemas/index.ts
alfonsograziano: Fair enough, will do it

apps/payments-api/routes/redirectToken/index.ts
alfonsograziano: It makes sense, I create a new route because it is related to the redirect and I didn't wanted to couple it too much to the transaction, but including in the transactions route is fair enough

apps/payments-api/app.ts
```diff
@@ -119,35 +115,5 @@ export async function build(opts?: FastifyServerOptions) {
     return getJWKSRoute(publicKey);
   });
 
-  // This is a test route that calls the integrator with a callback
-  app.get("/call-integrator", async () => {
```
alfonsograziano, line 123: Just removed a test API - we don't need it

peschina (approved):
LGTM

PR #1054, opened 2024-10-14
[Payments] Dynamic amount through encoded token

peschina:
### Ticket:

- [22479](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/22479)

### Description

If a token is present in query params in pay page url, override the amount in the payment request with the amount in the decoded token

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**
- [ ] **Additional tests**
- [ ] **Documentation**
- [ ] **Other**

### Checklist:

### Screenshots:

N/A

### DIFF
--- a/apps/integrator-api/app.ts
+++ b/apps/integrator-api/app.ts
@@ -13,7 +13,11 @@ import { fileURLToPath } from "url";
 import { dirname, join } from "path";
 import sensible from "@fastify/sensible";
 import schemaValidators from "./src/routes/schemas/validations";
-import apiAuthPlugin, { verifyJWT } from "api-auth";
+import apiAuthPlugin, {
+  getJWKSRoute,
+  readOrGenerateKeyPair,
+  verifyJWT,
+} from "api-auth";
 import { initializeErrorHandler } from "@ogcio/fastify-error-handler";
 import { initializeLoggingHooks } from "@ogcio/fastify-logging-wrapper";
 import healthCheck from "./src/routes/healthcheck";
@@ -87,6 +91,11 @@ export async function build(opts?: FastifyServerOptions) {
   app.register(healthCheck);
 
   app.register(routes, { prefix: "/api/v1" });
+
+  app.get("/.well-known/jwks.json", async () => {
+    const { publicKey } = await readOrGenerateKeyPair("integrator-api");
+    return getJWKSRoute(publicKey);
+  });
 
   // Test callback route to test communication between systems
   app.get("/callback", async (request, reply) => {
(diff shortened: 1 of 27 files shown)


apps/integrator-api/app.ts
```diff
@@ -88,6 +92,11 @@ export async function build(opts?: FastifyServerOptions) {
 
   app.register(routes, { prefix: "/api/v1" });
 
+  app.get("/.well-known/jwks.json", async () => {
+    const { publicKey } = await readOrGenerateKeyPair("integrator-api");
```
alfonsograziano, line 96: Thanks Sara! The next step here will then be to load this from KMS and Localstack.https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/22547

For this PR is fine to still generate the key at runtime, we will revisit it in the next task :)

apps/payments/app/[locale]/(embedded)/paymentRequest/pay/page.tsx
```diff
@@ -84,9 +97,8 @@ export default async function Page(props: Props) {
 
   const allowCustomAmount = details.allowCustomAmount;
 
-  let urlAmount = props.searchParams.amount
-    ? parseFloat(props.searchParams.amount)
-    : undefined;
+  let urlAmount = dynamicAmount?.amount ?? undefined;
```
Comment (alfonsograziano) [inline apps/payments/app/[locale]/(embedded)/paymentRequest/pay/page.tsx:100]: Shall we get rid of this urlAmount variable? Is just a little bit confusing for me, since is not coming anymore from a url

apps/payments/app/[locale]/(hosted)/paymentRequest/openbanking/page.tsx
```diff
@@ -32,8 +32,8 @@ async function getPaymentDetails(
   const realAmount = getRealAmount({
     amount: details.amount,
     customAmount,
-    amountOverride: amount,
-    allowAmountOverride: details.allowAmountOverride,
+    dynamicAmount: amount,
```
Comment (alfonsograziano) [inline apps/payments/app/[locale]/(hosted)/paymentRequest/openbanking/page.tsx:35]: Shouldn't we do the same token check on the actual payment pages? Otherwise, I can easily trick the system and still pay the amount I want. Maybe I'm missing something in the implementation

apps/payments/app/[locale]/(embedded)/paymentRequest/pay/page.tsx
Comment (peschina) [reply apps/payments/app/[locale]/(embedded)/paymentRequest/pay/page.tsx]: sure

apps/payments/app/[locale]/(hosted)/paymentRequest/openbanking/page.tsx
Comment (peschina) [reply apps/payments/app/[locale]/(hosted)/paymentRequest/openbanking/page.tsx]: good catch, I also made sure the correct amount is used when creating the transaction in case of custom amount chosen by the user

alfonsograziano (approved):
LGTM

PR #1072, opened 2024-10-17
[Integrator] Sign and verify with asymmetric key on KMS

peschina:
### Ticket:

- [22547](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/22547)

### Description

Sign and verify JWT used to communicate between services with asymmetric key created on KMS

Related [terraform PR](https://github.com/ogcio/life-events-terraform/pull/159)

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [ ] **New feature**
- [x] **Dev change**
- [ ] **Additional tests**
- [ ] **Documentation**
- [ ] **Other**

### Checklist:

### Screenshots:

N/A

### DIFF
--- a/apps/integrator-api/.env.sample
+++ b/apps/integrator-api/.env.sample
@@ -16,3 +16,8 @@ PAYMENTS_SERVICE_URL=http://localhost:8001/
 FORMS_SERVICE_URL=https://www.formsie.dev.gov.ie
 FORMS_SECRET_API_KEY=forms_secret_key
 FORMS_PUBLIC_API_KEY=form_public_key
+
+AWS_REGION=eu-west-1
+KMS_ENDPOINT=http://localhost:4566
+AWS_ACCESS_KEY_ID=123
+AWS_SECRET_ACCESS_KEY=123
--- a/apps/integrator-api/app.ts
+++ b/apps/integrator-api/app.ts
@@ -93,6 +93,7 @@ export async function build(opts?: FastifyServerOptions) {
     return getJWKS(keyAlias);
   });
 
+  // can be used to test
   app.get("/token", async (_, reply) => {
     const payload = { amount: 70 };
 
@@ -125,7 +126,6 @@ export async function build(opts?: FastifyServerOptions) {
         jwksUrl,
         issuer: "payments-api",
         audience: "integrator-api",
-        algorithm: "RS256",
       });
 
       // Send back the payload if the JWT verification is successful
--- a/apps/integrator-api/app.ts
+++ b/apps/integrator-api/app.ts
@@ -21,10 +21,10 @@ import journey from "./src/plugins/entities/journey/index.js";
 import journeyStepConnections from "./src/plugins/entities/journeyStepConnections";
 import journeySteps from "./src/plugins/entities/journeySteps";
 import run from "./src/plugins/entities/run";
+import { keyAlias } from "./src/utils/kms";
 
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = dirname(__filename);
-const keyAlias = "alias/life-events-integrator-api-key";
 
 dotenv.config();
 
--- a/apps/integrator-api/app.ts
+++ b/apps/integrator-api/app.ts
@@ -24,7 +24,7 @@ import run from "./src/plugins/entities/run";
 
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = dirname(__filename);
-const keyAlias = "alias/integrator-api-key";
+const keyAlias = "alias/life-events-integrator-api-key";
 
 dotenv.config();
 
--- a/apps/integrator-api/app.ts
+++ b/apps/integrator-api/app.ts
@@ -13,11 +13,7 @@ import { fileURLToPath } from "url";
 import { dirname, join } from "path";
 import sensible from "@fastify/sensible";
 import schemaValidators from "./src/routes/schemas/validations";
-import apiAuthPlugin, {
-  getJWKSRoute,
-  readOrGenerateKeyPair,
-  verifyJWT,
-} from "api-auth";
+import apiAuthPlugin, { createSignedJWT, getJWKS, verifyJWT } from "api-auth";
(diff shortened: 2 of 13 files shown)


apps/integrator-api/.env.sample
```diff
@@ -16,3 +16,8 @@ PAYMENTS_SERVICE_URL=http://localhost:8001/
 FORMS_SERVICE_URL=https://www.formsie.dev.gov.ie
 FORMS_SECRET_API_KEY=forms_secret_key
 FORMS_PUBLIC_API_KEY=form_public_key
+
+AWS_REGION=eu-west-1
+KMS_ENDPOINT=http://localhost:4566
+AWS_ACCESS_KEY_ID=123
+AWS_SECRET_ACCESS_KEY=123
```
peschina, line 23: These should be optional, needed as dummy values to connect to localstack km

apps/integrator-api/app.ts
```diff
@@ -28,6 +24,7 @@ import run from "./src/plugins/entities/run";
 
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = dirname(__filename);
+const keyAlias = "alias/integrator-api-key";
```
peschina, line 27: This value should stay the same across all environments. Any better place where to put it?

packages/api-auth/jwtService.ts
```diff
@@ -69,36 +49,45 @@ interface JWTOptions {
 /**
  * Creates and signs a JWT using the provided payload and private key.
  * @param payload - The JWT payload as an object.
- * @param privateKey - The private key to sign the token with.
+ * @param keyId - The key id or alias in KMS
  * @param options - Optional parameters for customizing the JWT creation (e.g., expiration, audience, issuer).
  */
 async function createSignedJWT(
```
peschina, line 55: The private key cannot be export it from KMS, so we need to create an encoded jwt manually, signing it directly on KMS

apps/integrator-api/app.ts
alfonsograziano: I think we can keep it here, but most probably I'd document it adding a couple of comments explaining what this is, why is matters and why we decided to keep the string here. We tend to avoid using magic words, but if will be the same across all the envs and will (hopefully) never change, we can keep it here - or if needed, in a utils/kms.ts file

apps/payments-api/app.ts
```diff
@@ -27,6 +27,8 @@ import rawbody from "fastify-raw-body";
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = dirname(__filename);
 
+const keyAlias = "alias/payments-api-key";
```
alfonsograziano, line 30: Same here, is fine to keep it here, but we should document what it is etc. Maybe creating it in a separated, well documented file might be a good idea. The less info we keep in the main app, the better it is

apps/payments-api/routes/transactions/index.ts
```diff
@@ -336,15 +336,15 @@ export default async function transactions(app: FastifyInstance) {
...
       );
 
-      const { privateKey } = await readOrGenerateKeyPair("payments-api");
       const jwt = await createSignedJWT(
         {
           userId: userId,
           transactionId: transactionDetails.transactionId,
         },
-        privateKey,
+        "alias/payments-api-key",
```
alfonsograziano, line 344: Yep, this confirms my idea :) We're fine with magic words, but we should avoid repeating them. Let's move this definition in a utils/kms.ts where we explain what is this, and then we can import the alias here

packages/api-auth/jwtService.ts
```diff
@@ -1,63 +1,43 @@
...
+
+  return function getKmsClient(): KMSClient {
+    if (kmsClient) return kmsClient;
+
+    const kmsConfig: KMSClientConfig = {
+      region: process.env.AWS_REGION,
+      endpoint: process.env.KMS_ENDPOINT,
+    };
+
+    if (process.env.AWS_ACCESS_KEY_ID && process.env.AWS_SECRET_ACCESS_KEY) {
```
alfonsograziano, line 28: In my experience with AWS, we don't have to pass these credentials as `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` are loaded and recognized automatically, maybe things changed, but I'd give it a try removing this if and the `.credentials` assignment

packages/api-auth/jwtService.ts
```diff
@@ -69,36 +49,45 @@ interface JWTOptions {
...
+  const command = new SignCommand(input);
+  const signResponse = await getKmsClient().send(command);
+
+  if (!signResponse.Signature) {
+    throw new Error("KMS did not return a signature. Signing failed.");
   }
-  return jwt.sign(pk);
+  const signature = Buffer.from(signResponse.Signature).toString("base64url");
+
+  return `${messageToSign}.${signature}`;
```
alfonsograziano, line 90: This looks a little bit different from the previous implementation. Is it still returning a JWT?

nnorbert (approved):
Code looks good, but I don't know kms, so I can't validate the flow.

packages/api-auth/jwtService.ts
alfonsograziano: Might another option to retrieve the key from the storage, and then sign? Similar to what you're doing on the `getJWKS` method. I've never used KMS so I'm not sure about the best approach. If this is the best approach, it might be helpful to add the links to some docs here

packages/api-auth/jwtService.ts
alfonsograziano: This actually solves one of my questions :D 
Can we please add this information in a comment within the code?

packages/api-auth/jwtService.ts
peschina: I commented above, these are needed only when running locally and relying on localstack. So the idea is to have them defined only locally

alfonsograziano: Uh makes sense thanks :)

alfonsograziano (approved):

PR #1076, opened 2024-10-17
22555 executor api transition step

nnorbert:
Integrator api: step transition


apps/integrator-api/src/libraries/integratorEngine/plugins/formPlugin.ts
```diff
@@ -6,4 +6,8 @@ export class FormPlugin implements IIntegratorPlugin {
     const { url } = stepData;
     return { url };
   }
+
+  public processResultData(data: any) {
```
alfonsograziano, line 10: At some point, we might want to ensure that data has a specific schema adding some validation. Is good for now though

apps/integrator-api/src/routes/executor/index.ts
```diff
@@ -225,6 +228,12 @@ export default async function executor(app: FastifyInstance) {
         throw app.httpErrors.unauthorized("Unauthorized!");
       }
 
+      if (!process.env.INTEGRATOR_URL) {
```
alfonsograziano, line 231: Instead of checking here, shall we check for the `INTEGRATOR_URL` variable in the fastify-env schema? I'm not a big fan of checking env vars at runtime, it can become really messy. It would be better if we just let the app crash if we're missing the env var

apps/integrator-api/src/routes/executor/index.ts
```diff
@@ -245,7 +255,12 @@ export default async function executor(app: FastifyInstance) {
       }
 
       const step = await app.journeySteps.getStepById(activeRunStep.stepId);
-      const engine = new IntegratorEngine(step.stepType);
+      const engine = new IntegratorEngine(
+        step.stepType,
+        journeyId,
+        runId,
+        process.env.INTEGRATOR_URL,
```
alfonsograziano, line 262: For now, we have 4 params, but I have the feeling that this might become bigger and bigger in the future, shall we refactor it to accept an object instead of 4 params?

apps/integrator-api/src/routes/executor/index.ts
```diff
@@ -256,4 +271,102 @@ export default async function executor(app: FastifyInstance) {
       reply.send(formatAPIResponse(result));
     },
   );
+
+  app.post<{
+    Reply: GenericResponse<JourneyStepExecutionResponse> | Error;
+    Body: TransitionJourneyStep;
+  }>(
+    "/transition",
```
alfonsograziano, line 279: This endpoint contains some interesting business logic, is it worth to add a few comments explaining what we're doing here and why?

apps/integrator/app/[locale]/(hosted)/journey/[journey]/callback/page.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { redirect, RedirectType } from "next/navigation";
+import { AuthenticationFactory } from "../../../../../../libraries/authentication-factory";
+
+type Props = {
+  params: {
+    locale: string;
+    journey: string;
+  };
+  searchParams: {
+    runId: string;
```
Comment (alfonsograziano) [inline apps/integrator/app/[locale]/(hosted)/journey/[journey]/callback/page.tsx:11]: Non-blocking: just something to keep in mind - maybe we can use the token to pass the runId as well, so that we're sure that we can trust the id. Maybe we can discuss it in another PR, shouldn't be a huge deal to refactor if we decide that is worth it

apps/integrator/app/[locale]/(hosted)/journey/[journey]/callback/page.tsx
```diff
@@ -0,0 +1,59 @@
...
+    return redirect("/admin/inactivePublicServant", RedirectType.replace);
+  }
+
+  if (isPublicServant) {
+    return redirect("/admin/journeys", RedirectType.replace);
+  }
+
+  const url = await transitionStep(journeyId, runId, searchParams);
+
+  if (!url) {
```
Comment (alfonsograziano) [inline apps/integrator/app/[locale]/(hosted)/journey/[journey]/callback/page.tsx:48]: At some point, we should probably create a `GenericErrorPage` and return that instead of this hardcoded string, not something we have to address now though

apps/integrator-api/src/routes/executor/index.ts
nnorbert: Done!

apps/integrator-api/src/routes/executor/index.ts
nnorbert: Done!

apps/integrator-api/src/routes/executor/index.ts
nnorbert: Done!

alfonsograziano (approved):
LGTM

PR #1107, opened 2024-10-28
[Payments]  Allow optional custom amount from JWT

peschina:
### Ticket:

- [22856](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/22856)
At the moment, if I pass a JWT, the "amount" field gets verified even if is not present in the JWT. We need to fix this behavior and check for the amount only if we have that field in the JWT.

### Description

Default to original amount if amount override option is enabled but custom amount is not present in JWT.

## Type

- [ ] **Dependency upgrade**
- [x] **Bug fix**
- [ ] **New feature**
- [ ] **Dev change**
- [ ] **Additional tests**
- [ ] **Documentation**
- [ ] **Other**

### Checklist:

### Screenshots:

N/A


alfonsograziano (approved):
LGTM

PR #1113, opened 2024-10-29
[Payments] Redirect to custom url from token

peschina:
### Ticket:

- [22854](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/22854)

### Description

Redirect to custom url taken from token if present upon payment completion, otherwise default to original payment request redirect url

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**
- [ ] **Additional tests**
- [ ] **Documentation**
- [ ] **Other**

### Checklist:

### Screenshots:

N/A


apps/payments/app/[locale]/(hosted)/paymentRequest/utils.ts
```diff
@@ -40,3 +40,21 @@ export async function getAmount({
     ? parseFloat(customAmount)
     : prDetails.amount;
 }
+
+export function getRedirectUrl({
+  tokenUrl,
+  originalUrl,
+}: {
+  tokenUrl: string;
```
Comment (alfonsograziano) [inline apps/payments/app/[locale]/(hosted)/paymentRequest/utils.ts:48]: Should this be `string | undefined`?

apps/payments/app/[locale]/(hosted)/paymentRequest/utils.ts
```diff
@@ -7,7 +7,7 @@ export async function getTokenPayload(token: string) {
   const paymentsApi = await AuthenticationFactory.getPaymentsClient();
   const { data: payload, error } = await paymentsApi.decodeToken({ token });
 
-  if (error) errorHandler(error);
+  if (error) console.log("error decoding token: ", error);
```
Comment (alfonsograziano) [inline apps/payments/app/[locale]/(hosted)/paymentRequest/utils.ts:10]: Is this a leftover, or do we want to keep the console.log?

alfonsograziano (dismissed):
LGTM - just left one comment

nnorbert (approved):
LGTM!

apps/payments/app/[locale]/(hosted)/paymentRequest/utils.ts
Comment (peschina) [reply apps/payments/app/[locale]/(hosted)/paymentRequest/utils.ts]: correct [f365c53](https://github.com/ogcio/life-events/pull/1113/commits/f365c5392993eff381660eb0ae13db134139cfdc)

apps/payments/app/[locale]/(hosted)/paymentRequest/utils.ts
Comment (peschina) [reply apps/payments/app/[locale]/(hosted)/paymentRequest/utils.ts:10]: In the ticket description it says we want to use the default redirect URL without interruption. So we need to ignore errors when decoding the token, but I thought we might want to log the error so to have visibility

Comment (alfonsograziano) [reply apps/payments/app/[locale]/(hosted)/paymentRequest/utils.ts:10]: Good catch! Thanks for explaining

alfonsograziano (approved):

PR #1122, opened 2024-10-30
use specific ubuntu version for running unit tests

David-Losinski:
### Ticket:

### Description
Unit tests are running using ubuntu-latest, we want to pin to a specific version as they are migrating the version ubuntu-latests points at and the newer version has had node 16 removed earlier in the month, which breaks building the design system package in the life events repo

https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [ ] **New feature**
- [ ] **Dev change**
- [ ] **Additional tests**
- [ ] **Documentation**
- [ ] **Other**

### Checklist:

### Screenshots:

N/A


relvao (approved):

azure_pipelines.yml
```diff
@@ -40,7 +40,7 @@ resources:
     - repository: pipeline-templates
       type: github
       name: ogcio/building-blocks-pipelines
-      ref: refs/tags/v0.2
+      ref: main
```
williamwd, line 43: Please refrain from using the branch name, as changes might break our pipelines without notice.

PR #1129, opened 2024-10-31
feat(integrator): submissions list page for public servants

nathanbirrell:
### Ticket:

- [22611](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/22611)

### Description

Adds list view of submissions for public servants.

At a high level, this adds:

* a new page to list all submissions
* a new endpoint to `getAllRuns` for public servants (NOTE: a "run" = a "submission"), with pagination
* a new filter on `getJourneys` to fetch journeys by array of `ids` (ie GET `/journeys?ids=123&ids=456`). **Note that this didn't end up being required, as I updated the getAllRuns query to resolve the journeyTitle instead, saving an extra trip to the DB.** I'll leave it in as it seems like a good API interface to have anyway. 

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**
- [ ] **Additional tests**
- [ ] **Documentation**
- [ ] **Other**

### Checklist:

### Screenshots:

Kapture 2024-10-31 at 13 02 05


apps/integrator/app/[locale]/(hosted)/admin/journeys/submissions/page.tsx
```diff
@@ -24,6 +24,45 @@ type Props = {
...
+
+  const journeyIds = new Set();
+
+  runsResult.data?.data.forEach((run) => journeyIds.add(run.journeyId));
+
+  const { data: journeysResult } = await integratorApi.getJourneys({
+    offset: 0,
+    limit: 10,
+    ids: Array.from(journeyIds),
+  });
```
Comment (nathanbirrell) [inline apps/integrator/app/[locale]/(hosted)/admin/journeys/submissions/page.tsx:45]: One improvement I'm looking at is removing this call. 

Since we already have the join on Run -> Journey in the `getAllRuns` call, we could avoid this extra trip to the DB by adding the Journey Name to the `getAllRuns` response.

apps/integrator/app/[locale]/(hosted)/admin/journeys/submissions/details/[submissionId]/page.tsx
```diff
@@ -65,6 +65,12 @@ export default async ({ params: { locale, submissionId } }: Props) => {
     return notFound();
   }
 
+  // TODO: lookup payment step, based on what? something in step.data?
+  const paymentStep = submission?.steps[0];
+  // TODO: confirm data.transactionId key
+  // @ts-expect-error TODO: type data properly
+  let transactionId = paymentStep?.data?.transactionId;
```
Comment (nathanbirrell) [inline apps/integrator/app/[locale]/(hosted)/admin/journeys/submissions/details/[submissionId]/page.tsx:72]: Left a TODO because this will need to be addressed once the data structure for `data` is decided.

apps/intg-api/src/plugins/entities/journey/repo.ts
```diff
@@ -31,11 +33,32 @@ export class JourneyRepo {
...
+
+    if (filter?.ids) {
+      // NOTE: when only one ID is passed, it is a string, but when multiple IDs are passed, it is an array
+      const ids = typeof filter.ids === "string" ? [filter.ids] : filter.ids;
+
+      // TODO: is there a safer way to filter on an array of IDs? The concern being
+      // that as we add more arguments to filter on, we need to be careful about the
+      // order of the arguments
+      queryString += `AND id = ANY($4::uuid[]) `;
+      queryValues.push(ids);
```
nathanbirrell, line 53: **Question:** interested to hear ideas on improving this, or if there's an example of a better pattern in this app?

apps/intg-api/src/plugins/entities/run/index.ts
```diff
@@ -76,6 +76,52 @@ const buildGetRunById =
...
+      result = await repo.getRuns(organizationId, pagination);
+    } catch (err) {
+      log.error((err as Error).message);
+    }
+
+    if (!result?.rowCount) {
+      throw httpErrors.notFound("The requested run was not found");
+    }
+
+    // TODO: do we need to check Civil Servants can only access their orgs submissions?
```
nathanbirrell, line 97: @alfonsograziano maybe you can answer this one?

apps/intg-api/src/routes/schemas/index.ts
```diff
@@ -43,6 +43,16 @@ export const PaginationLinks = Type.Object({
 });
 export type PaginationLinks = Static<typeof PaginationLinks>;
 
+/**
+ * Generic filter options
+ */
+
+export const FilterParams = Type.Object({
+  // Union because if only one ID is provided in the query, it is a string, but if multiple IDs are provided, it is an array
+  ids: Type.Optional(Type.Union([Type.String(), Type.Array(Type.String())])),
```
nathanbirrell, line 52: **Note:** interesting discovery here. Not sure if it's Fastify or ajv but the `ids` array filter becomes a `string` when only **one id is passed** at runtime. Thanks @nnorbert for this solution.

apps/integrator/app/[locale]/(hosted)/admin/journeys/submissions/page.tsx
Comment (nathanbirrell) [reply apps/integrator/app/[locale]/(hosted)/admin/journeys/submissions/page.tsx]: This is now done

apps/integrator/app/[locale]/(hosted)/admin/journeys/submissions/details/[submissionId]/page.tsx
```diff
@@ -154,7 +160,17 @@ export default async ({ params: { locale, submissionId } }: Props) => {
               <dt className="govie-summary-list__key">
                 {t("paymentDetails.submissionLink")}
               </dt>
-              <dt className="govie-summary-list__value">Todo</dt>
+              <dt className="govie-summary-list__value">
+                {transactionId ? (
+                  <a
```
Comment (nnorbert) [inline apps/integrator/app/[locale]/(hosted)/admin/journeys/submissions/details/[submissionId]/page.tsx:165]: Instead of the standard HTML anchor tag, we use Link from next

```
import Link from "next/link";
...
<Link className="govie-link" href="url">
  text
 </Link>
```

Comment (nathanbirrell) [reply apps/integrator/app/[locale]/(hosted)/admin/journeys/submissions/details/[submissionId]/page.tsx:165]: oops, good spotting! forgot to change this

apps/intg-api/src/plugins/entities/journey/repo.ts
nnorbert: I also used the same approach, slightly different implementation, instead of concatenating the query strings, I stored them into an array and rendered them with Array.join(). Since we don't use any query builder, this seems to be the most straightforward solution.

apps/intg-api/src/plugins/entities/run/index.ts
nnorbert: No, not in entity level, that check must be done in the HTTP layer.

nnorbert (approved):
LGTM

PR #1133, opened 2024-11-04
Feat/22930 Message widget setup

Thurfjell:
### Ticket:

- [22930](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/22930)

### Description

Set up the messaging widget boilerplate persistence by list selectable templates. 

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [x] **Dev change**
- [ ] **Additional tests**
- [ ] **Documentation**
- [ ] **Other**

### Checklist:

### Screenshots:

N/A


alfonsograziano (approved):
LGTM

apps/integrator/libraries/messaging-authentication-factory.ts
```diff
@@ -16,7 +18,32 @@ export class MessagingAuthenticationFactory {
     return token;
   }
 
-  static async getMessagingClient(): Promise<Messaging> {
+  static async getMessagingE2EClient(): Promise<Messaging> {
```
nnorbert, line 21: The naming is confusing here. This is not used in the E2E tests and most probably it is not using the E2E logo application, does it? I think this should be only getMessagingClient.

Thurfjell: How is it confusing? The method isn't composing a result, but internally depends on things instead.. I'm quite. positive we don't want to use the notion of `end to end`  being defacto sdk client for general messaging. Especially since, at least in my reality, we use the term `end to end` when talking about testing. If anything, this should be removed I don't understand why it's here in the first place, I can't find any end to end tests in this specific application.

PR #1146, opened 2024-11-06
Feat(22931) Messaging plugin for integrator

Thurfjell:
### Ticket:

- [22931](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/22931)

### Description

Import active journey user to messaging application
Send message based on template from integrator, as public and email.

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**
- [ ] **Additional tests**
- [ ] **Documentation**
- [ ] **Other**

### Checklist:

### Screenshots:

N/A


apps/intg-api/src/routes/schemas/index.ts
```diff
@@ -118,6 +118,7 @@ export const PaymentStepData = Type.Object({
 export const MessagingStepData = Type.Object({
   templateId: Type.String(),
   title: Type.String(),
+  organisationId: Type.String(),
```
alfonsograziano, line 121: Why do we need to save the orgnizationId here? Just asking since in Payments we're not saving it. Are we missing something maybe?

Thurfjell: When sending a message, we need to know where it was sent from. I think it's a hard requirement to acquire a messaging sdk as well?

apps/intg-api/src/routes/executor/index.ts
```diff
@@ -344,6 +344,7 @@ export default async function executor(app: FastifyInstance) {
         journeyId,
         runId,
         host: process.env.INTEGRATOR_URL!,
+        logtoUserId: userId,
```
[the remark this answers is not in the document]
Thurfjell: I'd rather be more explicit which ids we pass around. We don't really have any determined centralised `user id` (profile, <domain>, Logto). I've dealt with 3 different user ids in one and the same function. 🙃

nnorbert (dismissed):
LGTM!

nnorbert (approved):

PR #1155, opened 2024-11-07
Design fixes

nnorbert:
Design fixes


apps/integrator/app/components/radioButton.tsx
```diff
@@ -0,0 +1,42 @@
+"use client";
```
Thurfjell, line 1: Why?

Thurfjell: Just use default checked prop and it's server compatible.

apps/integrator/app/styles/globals.scss
Thurfjell, line 1: 🥲

Thurfjell (approved):

apps/integrator/app/components/InputField.tsx
```diff
@@ -30,37 +30,47 @@ export default ({
...
-          {error}
-        </p>
-      )}
-      <div className="govie-input__wrapper">
-        {prefix && (
-          <div aria-hidden="true" className="govie-input__prefix">
-            {prefix}
-          </div>
+    <>
+      <style jsx>{`
```
alfonsograziano, line 34: Non-blocking: How about using CSS Modules? Or even an inline style since you're using this class only in this case

apps/integrator/app/components/radioButton.tsx
alfonsograziano, line 1: Just something minor, if we want to follow the usual naming strategy, the file name should be `RadioButton` (with the capital R)

apps/integrator/app/styles/globals.scss
```diff
@@ -120,3 +120,127 @@ $breakpoint-xxl: 1400px;
   align-items: center;
   justify-content: center;
 }
+
+// Radio button medium size related styles - temporary solution
```
alfonsograziano, line 124: Please as this is a temporary solution, let's add a `TODO` statement, so that we can reference it easily later and fix the temporary hack

alfonsograziano (dismissed):
LGTM apart from some minor comments and suggestions

apps/integrator/app/styles/globals.scss
Thurfjell: I think the solution to this beautiful adjustment is actually using the components from the new design system's react component library, so ie remove all files and old design system imports 😄

apps/integrator/app/components/radioButton.tsx
nnorbert: It is changed :)

apps/integrator/app/components/InputField.tsx
nnorbert: Done

apps/integrator/app/components/radioButton.tsx
nnorbert: Done

apps/integrator/app/styles/globals.scss
nnorbert: Done

alfonsograziano (approved):
LGTM

PR #1170, opened 2024-11-12
Add QR code within the PDF generation

alfonsograziano:
We need to generate a QR code. The layout, the verbiage and the content of the PDF are not final and need to go through a design review.

Screenshot 2024-11-12 at 08 12 07


Thurfjell (approved):

apps/intg-api/src/libraries/pdfBuilder/index.ts
```diff
@@ -130,6 +144,19 @@ const generateSubmissionPDF = async (
                 <dt class="govie-summary-list__value">${details.completeDate}</dt>
               </div>
             </dl>
+            <div style="display: flex; align-items: center; margin-top: 30px;">
+            <div>
+              <a href="${qrCodeDataUrl}">
+                <img src="${qrCodeDataUrl}" alt="QR Code" style="width: 150px; height: 150px;" />
```
nnorbert, line 150: Where is this URL pointing to?

alfonsograziano: when clicked, it should simply open an image of the QR code. This is not final though, we will need to refine the layout, wording etc in an upcoming PR

nnorbert (dismissed):
LGTM!

PR #1185, opened 2024-11-14
23158 required journey steps

nnorbert:
Validate journey steps and allow required fields.


apps/integrator/libraries/journeyEditor/widgets/index.ts
```diff
@@ -8,7 +8,7 @@ import PaymentWidget from "./paymentWidget";
 import TitleWidget from "./titleWidget";
 
 export interface JourneyStepWidget {
-  getForm: (data: any) => Promise<JSX.Element>;
+  getFormData: (data: any) => Promise<any>;
```
peschina, line 11: Do you think we could avoid any here?

apps/integrator/app/[locale]/(hosted)/admin/journeys/create/page.tsx
```diff
@@ -40,44 +56,42 @@ export default async ({ params: { locale } }: Props) => {
...
-    );
+    let journeyResult;
+    try {
+      journeyResult = await editor.createJourney(title);
+    } catch (error) {
+      /**
+       * Journey and steps creation failed, logging the reason
+       * and redirecting the user to the error page.
+       */
+      console.error(error);
```
Comment (peschina) [inline apps/integrator/app/[locale]/(hosted)/admin/journeys/create/page.tsx:68]:
```suggestion
      console.error("Journey and steps creation failed: ", error);
```

apps/intg-api/src/routes/schemas/index.ts
```diff
@@ -131,14 +136,14 @@ export const StepData = Type.Union([
   PaymentStepData,
   MessagingStepData,
   CompleteStepData,
+  Type.Object({}),
```
peschina, line 139: Why this? Shouldn't it be `TitleStepData`?

apps/intg-api/src/routes/schemas/index.ts
```diff
@@ -151,10 +156,20 @@ export const CreateJourneyStep = Type.Pick(JourneyStep, [
...
 ]);
 
-export const UpdateJourneyStep = Type.Pick(JourneyStep, [
-  "stepData",
-  "configured",
-]);
+export const UpdateJourneyStep = Type.Object({
+  stepData: Type.Union(
+    [
+      Type.Object({}),
```
peschina, line 162: same as above

peschina (dismissed):
Looks good, some minor comments

apps/intg-api/src/routes/schemas/index.ts
nnorbert: This is the tricky part. The steps will be created with no data. Without this, the validation fails. I will try one more thing, because I observed that I am setting an empty object. Maybe, if I would set undefined, and set the StepData optional, maybe that would work.

apps/integrator/libraries/journeyEditor/widgets/index.ts
nnorbert: Yes, this is PluginData, I just forgot to change it.

nnorbert: Done

apps/intg-api/src/routes/schemas/index.ts
nnorbert: Does not work with Optional.

peschina (approved):
LGTM

PR #1190, opened 2024-11-15
ux-adjustments

nnorbert:
UX Adjustments


alfonsograziano (approved):
LGTM - at some point we will need to remove some dead code (like the Worldpay provider setup) but this is something we can address in other PRs

PR #1198, opened 2024-11-19
feat(23377): custom not found and error pages

nnorbert:
Custom error and page not fund pages added to the integrator app.
Updated the language switch logic in payments' not found page.


alfonsograziano (approved):
LGTM

PR #1221, opened 2024-11-25
22966 move journey api into bb sdk

nnorbert:
Move Journey API into the Building Blocks SDK package.
Only Journey service was updated, Payments and Dashboard services are still using the Journey API calls from the old package. That will be updated once the new Journey API package was published.


apps/journey/libraries/building-block-sdk.ts
```diff
@@ -0,0 +1,75 @@
...
+  const res = await fetch(
+    new URL(
+      path,
+      process.env.NEXT_PUBLIC_JOURNEY_SERVICE_ENTRY_POINT as string,
+    ),
+    { headers: { cookie: cookieHeader } },
+  );
+  const { token } = await res.json();
+  return token;
+}
```
peschina, line 75: Can you please add EOF?

peschina (approved):
LGTM, one minor comment

apps/journey/libraries/building-block-sdk.ts
nnorbert: Done!

alfonsograziano (dismissed):
LGTM

peschina (approved):

PR #1224, opened 2024-11-25
First unit tests for JB-API

alfonsograziano:
### Description

Setup of vite + a few unit tests

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [ ] **New feature**
- [ ] **Dev change**
- [X] **Additional tests**
- [ ] **Documentation**
- [ ] **Other**


apps/journey-api/src/test/engine/plugins/formPlugin.test.ts
```diff
@@ -0,0 +1,71 @@
...
+// Mock module
+vi.mock("api-auth", () => ({
+  createSignedJWT: vi.fn(async () => {
+    return "mocked-signed-jwt";
+  }),
+}));
+
+describe("Form plugin", () => {
+  it("should generate a URL with a signed JWT", async () => {
+    const { FormPlugin } = await import(
```
nnorbert, line 14: Why is this declared in the test case instead of the test suit level?

nnorbert: Why is this imported here instead of importing it in the traditional way? Like in the Messaging test case.

apps/journey-api/src/test/repositories/journeyStepsRepo.ts
```diff
@@ -0,0 +1,110 @@
+import { QueryResult } from "pg";
+import {
+  CreateJourneyStepDO,
+  JourneyStepDO,
+  UpdateJourneyStepDO,
+} from "../../plugins/entities/journeySteps/types";
+
+export class JourneyStepsRepo {
```
nnorbert, line 8: Please, rename the file and the class to be evident that this is a mock.

```
ex: JourneyStepMockRepo
```

apps/journey-api/src/test/engine/plugins/messagingPlugin.test.ts
```diff
@@ -0,0 +1,83 @@
...
+    journeyId: "journey123",
+    runId: "run456",
+    host: "https://jb-host.com",
+    journeyTitle: "Test Journey",
+    createdAt: "now",
+    logtoUserId: "user1",
+  };
+
+  it("should execute successfully and return a callback URL", async () => {
+    const messagePlugin = new MessagePlugin();
```
nnorbert, line 75: The instance can be moved at suite level. If a new instance is needed per test case, we can use `beforeEach`.

apps/journey-api/src/test/engine/plugins/paymentsPlugin.test.ts
```diff
@@ -0,0 +1,77 @@
+import { describe, it, expect, vi } from "vitest";
```
nnorbert, line 1: All the previous comments are valid here as well.

apps/journey-api/src/test/engine/plugins/formPlugin.test.ts
alfonsograziano: Good catch. Done :)

apps/journey-api/src/test/repositories/journeyStepsRepo.ts
alfonsograziano: Done, good catch :D

nnorbert (approved):

PR #1248, opened 2024-11-29
feat(22393): implement jwks endpoint

nnorbert:
Implement jwks endpoint


peschina (approved):
LGTM

apps/journey-api/src/libraries/integratorEngine/plugins/formPlugin.ts
```diff
@@ -50,8 +50,8 @@ export class FormPlugin implements IIntegratorPlugin {
     } as FormRunStepDataDO;
   }
 
-  //TODO: Add here the Forms Backend URL
   public getJwksUrl() {
-    return "/.well-known/jwks.json";
+    const url = new URL("/.well-known/jwks.json", process.env.FORMS_SERVICE_URL);
```
alfonsograziano, line 54: This endpoint is returning me 404. @nnorbert before merging is worth checking with Florian what the final endpoint will be

nnorbert: I have checked with Florian, he confirmed that the endpoint will be this, but the feature is not done yet.

alfonsograziano (approved):
LGTM

PR #1252, opened 2024-12-02
feat(22966): replace old integrator and sdk usages

nnorbert:
Replace old SDK usages


apps/dashboard/utils/authentication-factory.ts
```diff
@@ -243,8 +244,11 @@ class SDKFactory {
         featureFlags: {
           baseUrl: this.featureFlagApiHostUrl,
         },
+        journey: {
+          baseUrl: this.featureFlagApiHostUrl,
+        }
       },
-    })
+    }) as BuildingBlocksSDK
```
Thurfjell, line 251: Suggestion: Remove type assertion and set the type (probably like Pick services from returntype of getbuildingblockssdk or something).

nnorbert: Done!

Thurfjell (approved):

apps/journey-api/.env.sample
peschina, line 1: We will need to remove these environment variables in dev too

apps/payments-api/.env.sample
peschina, line 1: Same here, needs to be added in dev

peschina (approved):
LGTM. We need to update env vars in dev before merging

PR #1264, opened 2024-12-04
longer timeout

SamSalvatico:
Ticket: [22380](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_boards/board/t/Messaging/Stories?System.AreaPath=Messaging&workitem=22830)

First try setting a higher timeout (just an attempt, I would like to not spend too much effort on this given that we will need to change a lot of code with the new user onboarding)


Thurfjell (approved):
let's try

PR #1281, opened 2024-12-05
feat(23750): consume custom amount

nnorbert:
Get custom amount from Forms submission data and send it to Payments.


apps/journey-api/src/libraries/integratorEngine/index.ts
```diff
@@ -42,8 +42,8 @@ class IntegratorEngine {
     this.logtoUserId = props.logtoUserId;
   }
 
-  public executeStep(data: StepDataDO): Promise<{ url: string }> {
-    return this.plugin.execute(data, {
+  public executeStep(data: StepDataDO, submissionData: SubmissionData[]): Promise<{ url: string }> {
```
alfonsograziano, line 45: shall we rename `submissionData` in `runData` for coherence? And maybe, `runData` might be an object with the entire context of the current run (as we might need run information, with a field for the `runSteps`. Might be an idea for you?

apps/journey-api/src/libraries/integratorEngine/plugins/paymentPlugin.ts
```diff
@@ -8,8 +8,21 @@ import { PaymentRunStepData } from "../../../routes/schemas";
 
 export class PaymentPlugin implements IIntegratorPlugin {
   private readonly paymentLink = "/paymentRequest/pay";
+
+  private getCustomAmount(submissionData: SubmissionData[]) {
+    const formSubmission = submissionData.find((submission) =>
```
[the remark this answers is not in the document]
nnorbert: Definitely, if you have to get a specific step-related value, you must know the reference to it. We can have a separate investigation, as you mentioned, sure.

apps/journey-api/src/libraries/integratorEngine/index.ts
nnorbert: Done

alfonsograziano (approved):
LGTM

PR #1334, opened 2024-12-16
[Journey] E2E test - journey creation with payment

peschina:
### Ticket:

- [23900](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/23900)

### Description

Add E2E test for journey creation with form and payment steps

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [ ] **New feature**
- [x] **Dev change**
- [ ] **Additional tests**
- [ ] **Documentation**
- [ ] **Other**

### Checklist:

### Screenshots:

N/A


apps/payments-api/routes/paymentRequests/index.ts
```diff
@@ -226,7 +227,7 @@ export default async function paymentRequests(app: FastifyInstance) {
       },
     },
     async (request, reply) => {
-      const userId = request.userData?.userId;
+      const userId = request.userData?.isM2MApplication ? deriveUserId(request.userData?.userId) : request.userData?.userId;
```
alfonsograziano, line 230: The PR looks good, the only thing that makes me think actually is that we're using M2M to impersonate the actions of a test user. Have we considered to use a test user with impersonation instead? This would lead to: 

- no need for an M2M application 
- no need to derive a fake id 
- enhanced security, as we can add only the user to the test organization and cannot act outside of that one

What do you think? I'm not against this solution, I was just thinking that maybe in the context of the E2E, impersonation might be another valuable option - even though I never actually used this feature

peschina: I think the M2M approach is the recommended one in case of service to service authentication. The M2M token is scoped, so it's not valid for other organisations, and gives us a nice separation of concerns (we don't put identities in the mix).
With the impersonation approach we would need access to client id and client secret of payments, an application that is not in the scope of our tests. This would be quite risky, as it's another set of credentials to store. We would also violate the least privilege principle I think, as our test user might have more privileges than what we need.

peschina: On another slightly different note, I am a bit concerned that my solution for the derived id might return an id that clashes with another user id, since I'm truncating it to 12 char. It is fine for the now since we're running these tests only locally, but I will discuss with Norbert for a more robust solution

alfonsograziano (approved):

PR #1427, opened 2025-02-06
Test(messaging-api): add some tests

SamSalvatico:
### Description

Added some tests
updated packages


apps/messaging-api/src/test/services/messages/message-service.test.ts
```diff
@@ -13,6 +13,12 @@ import { getMockBaseLogger } from "../../test-server-builder.js";
 
 let pool: Pool;
 
+let schedulerWorks = true;
+
+afterEach(() => {
+  schedulerWorks = true;
+});
```
Thurfjell, line 20: Tests never run in parallel right?

What you could do, is define the scheduleTasks function resolve in each it clause instead, and explicitly for each case determine the result.

SamSalvatico: Always serial
I need to check docs, with tap you can only mock once per file , anyway will give a look tomorrow, thanks

SamSalvatico: From the docs

> The call to vi.mock is hoisted, so it doesn't matter where you call it. It will always be executed before all imports. If you need to reference some variables outside of its scope, you can define them inside [vi.hoisted](https://vitest.dev/api/vi.html#vi-hoisted) and reference them inside vi.mock.

Then I can't mock it multiple times
https://vitest.dev/api/vi.html#vi-mock

Thurfjell: Aight should be fine. Im fairly certain you can assign things in each test. dynamic import (you mock at top) and assign via mockResolve/Reject. Should be some examples eg. in profile and dashboard.

Thurfjell (approved):

PR #1462, opened 2025-02-28
25561/sendamessage new recipients page

Thurfjell:
### Ticket:

- [25561](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/25561)

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [ ] **New feature**
- [x] **Dev change**
- [ ] **Additional tests**
- [ ] **Documentation**
- [ ] **Other**

### Checklist:

### Screenshots:

N/A


apps/messaging/components/send-message/Recipients.tsx
```diff
@@ -1,167 +1,476 @@
...
+  const handleAddRecipient = async (e: React.FormEvent<HTMLFormElement>) => {
+    e.preventDefault()
+    const formData = new FormData(e.currentTarget)
+    const firstName = formData.get("firstName") as string
+    const surname = formData.get("surname") as string
+    const email = formData.get("email") as string
+
+    setIsAdding(true)
+
+    const existanceCheck = await getRecipients({
```
SamSalvatico, line 184: Just to let you know, you can also leave as is if you want
To find and return  a single recipient you can also use profileClient.findProfile

apps/messaging/app/[locale]/admin/send-a-message/addRecipient.ts
```diff
@@ -0,0 +1,22 @@
+"use server"
+import { AuthenticationFactory } from "utils/authentication-factory"
+
+export async function addRecipient(
```
Comment (matteolc) [inline apps/messaging/app/[locale]/admin/send-a-message/addRecipient.ts:4]: could we use hash parameters instead of positional ones?

apps/messaging/app/[locale]/admin/send-a-message/getRecipients.ts
```diff
@@ -4,18 +4,28 @@ import { AuthenticationFactory } from "utils/authentication-factory"
...
+    lastName?: string
+    email?: string
+  }
 }) {
   const profile = await AuthenticationFactory.getProfileClient()
   const response = await profile.listProfiles({
     limit: queryParams.limit ? String(queryParams.limit) : undefined,
     offset: queryParams.offset ? String(queryParams.offset) : undefined,
     search: queryParams.search,
+    email: queryParams.email,
```
Comment (matteolc) [inline apps/messaging/app/[locale]/admin/send-a-message/getRecipients.ts:21]: How about?

```
const { limit, offset, ...otherProps } = queryParams
const response = await profile.listProfiles({
  limit: limit ? String(limit) : undefined,
  offset: offset ? String(offset) : undefined,
  ...otherProps
})
```

Comment (Thurfjell) [reply apps/messaging/app/[locale]/admin/send-a-message/getRecipients.ts:21]: What is the upside to spread, rather than be explicit with this few parameters? We save 3 LOC and obfuscate parameters.

apps/messaging/app/[locale]/admin/send-a-message/addRecipient.ts
Comment (Thurfjell) [reply apps/messaging/app/[locale]/admin/send-a-message/addRecipient.ts:4]: What does this mean?

Comment (Thurfjell) [reply apps/messaging/app/[locale]/admin/send-a-message/addRecipient.ts:4]: Ah, object param? 🙃

apps/messaging/app/[locale]/admin/send-a-message/getRecipients.ts
Comment (matteolc) [reply apps/messaging/app/[locale]/admin/send-a-message/getRecipients.ts:21]: Not really obfuscated as they are typed in the function arguments. If you add props you don't need to type them everywhere explicitly (same as for the previous comments, you could swallow these props, we know what type they are/must be)

matteolc (approved):

PR #1479, opened 2025-03-12
25419 - Message api events status query param

Thurfjell:
### Ticket:

- [25419](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/25419)

### Description

Added optional status Delivered, Scheduled, Opened and Failed query parameter to get message events endpoint.

Changed the default date range from [now - 1 year] to [now - 30 days].

Note: Very slow on large data sets

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [x] **Dev change**
- [ ] **Additional tests**
- [ ] **Documentation**
- [ ] **Other**

### Checklist:

### Screenshots:

N/A


apps/messaging-api/src/services/message-events/message-event-service.ts
```diff
@@ -21,14 +23,33 @@ export async function listMessageEvents(params: {
     search = "",
     dateFrom,
     dateTo,
+    status
   } = params;
 
+
+  let eventsStatusFilter = ""
```
SamSalvatico, line 30: what about using a switch?

apps/messaging-api/src/services/message-events/message-event-service.ts
```diff
@@ -37,8 +58,9 @@ export async function listMessageEvents(params: {
             AND mel.data->>'receiverFullName' ILIKE $2
         )
       )
+      ${messagesStatusFilter}
       AND m.scheduled_at::DATE BETWEEN 
-        COALESCE($3::DATE, (NOW() AT TIME ZONE 'UTC')::DATE - INTERVAL '365 days')
+        COALESCE($3::DATE, (NOW() AT TIME ZONE 'UTC')::DATE - INTERVAL '30 days')
```
SamSalvatico, line 63: why did you move to 30? Just asking

apps/messaging-api/src/services/message-events/message-event-service.ts
```diff
@@ -21,14 +23,33 @@ export async function listMessageEvents(params: {
...
+    eventsStatusFilter = "AND l.event_status='successful' AND l.event_type = 'message_schedule'"
+  }
+
   // Separate count query first
   const countQuery = await pool.query<{ count: number }>(
     `
     SELECT COUNT(DISTINCT m.id) as "count"
     FROM messages m
     WHERE m.organisation_id = $1
+      AND EXISTS (SELECT 1 FROM messaging_event_logs l
```
SamSalvatico, line 48: I didn't test this, but are you sure that this is not querying for the "latest event that has this status" and not for "check if the latest event has this status"?

E.g. 
Querying for failed events
Two events
1 is failed
2 is success (that is the latest inserted)

This where clause selects the latest event that is failed, that in this case is the 1, but we just want to query on the latest

apps/messaging-api/src/test/services/message-events/message-event-service.test.ts
```diff
@@ -554,4 +558,124 @@ describe("Message Event Service - listEvents", () => {
...
+    const events = await listMessageEvents({
+      pagination: { offset: "0", limit: "100" },
+      pool,
+      organizationId: organizationId,
+      status: "opened",
+    });
+
+    expect(events.totalCount).toEqual(1)
+    expect(events.data.at(0)?.messageId).toEqual(message1.id)
+  })
```
SamSalvatico, line 680: please add a test where the latest inserted event has status A, the previous one has status B and, searching for B, the event must not be returned, given that we only search for latest event

apps/messaging-api/src/services/message-events/message-event-service.ts
Thurfjell: See last comment on the ticket.

apps/messaging-api/src/services/message-events/message-event-service.ts
Thurfjell: I'll test and come back to you 👍

apps/messaging-api/src/services/message-events/message-event-service.ts
Thurfjell: No strong opinions on such small set. Would you prefer one?

SamSalvatico: yes, thank you

SamSalvatico (approved):

PR #1490, opened 2025-03-13
Feat(messaging): 25422 - redirect to hit page after login

SamSalvatico:
### Ticket:

- [25422](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/25422)

### Description

Now when a user tries to hit a page but is not logged in, the hit path is stored in a cookie, so to be used after Logto redirect to redirect to the desired page

I "took inspiration" 😁  from the PR from the payments team, the only additional feature I've added is the possibility to pass query parameters to the path we want to redirect to

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [ ] **New feature**
- [x] **Dev change**
- [ ] **Additional tests**
- [ ] **Documentation**
- [ ] **Other**


apps/messaging/utils/logto-config.ts
```diff
@@ -221,3 +234,16 @@ export const getSignInConfiguration = () => {
...
+  const currentSearch = headers().get(CustomHeaders.Search) ?? "";
+  let redirectPath = currentPath;
+  if(currentSearch.trim().length > 0) {
+    // we need to encode the search params to avoid them to be
+    // interpreted as query params of the actual path
+    redirectPath = `${redirectPath}${encodeURIComponent(currentSearch.trim())}`
+  }
+  const queryParams = `${PostLoginSearchParams.LoginUrl}=${routes.logtoLogin.url}&${PostLoginSearchParams.PostLoginRedirectPath}=${redirectPath}`;
+  return `${routes.preLogin.url}?${queryParams}`;
+};
```
Thurfjell, line 249: Could also use the URL api?

Thurfjell (approved):

apps/messaging/utils/logto-config.ts
SamSalvatico: I've tried with it, but given that `routes.preLogin.url` is not a full url, but only a path (`/pre-login`), it doesn't work

Thank you for the hint

PR #1506, opened 2025-03-20
26370 - Signout all apps, messaging dashboard dashboard-admin

Thurfjell:
### Ticket:

- [26370](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/26370)

### Description

Delete all cookies before signing out. 

In case of messaging, also set the logout redirect url prefixed with /admin, so that the sign in experience will pick up correct social connector.

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [ ] **Dev change**
- [ ] **Additional tests**
- [ ] **Documentation**
- [ ] **Other**

### Checklist:

### Screenshots:

N/A


apps/messaging/app/[locale]/signout/route.ts
```diff
@@ -1,8 +1,20 @@
 import { getCachedConfig } from "@/utils/env-config"
 import { AuthenticationRoutes } from "@ogcio/nextjs-auth"
+import { cookies, headers } from "next/headers"
 import { getSignInConfiguration } from "utils/logto-config"
 
 export async function GET() {
-  const { baseUrl } = getCachedConfig()()
-  await AuthenticationRoutes.logout(getSignInConfiguration(), baseUrl)
+  const referer = headers().get("referer")
```
Comment (matteolc) [inline apps/messaging/app/[locale]/signout/route.ts:7]: I think it's more like `referrer`

PR #1511, opened 2025-03-25
26770 - messaging-api role based message fetch

Thurfjell:
### Ticket:

- [26770](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/26770)

### Description

New reponse type and extendable getPartialMessage function in services. 

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [ ] **New feature**
- [ ] **Dev change**
- [ ] **Additional tests**
- [ ] **Documentation**
- [ ] **Other**

### Checklist:

### Screenshots:

N/A


apps/messaging-api/src/routes/messages/index.ts
```diff
@@ -79,16 +80,44 @@ export default async function messages(app: FastifyInstance) {
     "/:messageId",
     {
       preValidation: (req, res) =>
-        app.checkPermissions(req, res, [Permissions.MessageSelf.Read]),
+        app.checkPermissions(req, res, [
+          Permissions.MessageSelf.Read,
+          "messaging:message.m2m.limited:read",
```
SamSalvatico, line 85: I would avoid to make it specific for m2m, in future we could have a "human" role with a permission like this
`messaging:message.limited:read`

Or maybe `partial` instead of `limited`?

SamSalvatico: And also set it as constant in the Permissions namespace

apps/messaging-api/src/routes/messages/index.ts
```diff
@@ -79,16 +80,44 @@ export default async function messages(app: FastifyInstance) {
...
+        ]),
       schema: GetMessageReqSchema,
     },
     async function getMessageHandler(request, _reply) {
-      return {
-        data: await getMessage({
+      let data: GetMessageResponse["data"] | undefined;
+
+      if (request.userData?.isM2MApplication) {
```
SamSalvatico, line 92: would remove the is m2mApplication check

apps/messaging-api/src/services/messages/message-service.ts
```diff
@@ -39,6 +39,50 @@ export async function listMessages(params: {
   });
 }
 
+export async function getPartialMessage<
+  T extends keyof Pick<ReadMessage, "organisationId" | "recipientUserId">,
```
SamSalvatico, line 43: I still vote for the strict type without `pickedFields` param

I think @matteolc vote would decide this

apps/messaging-api/src/types/messages.ts
SamSalvatico, line 1: Please use a type like ReadPartialMessageSchema to define this
`Type.Pick(ReadMessageSchema, ["recipientUserId", "organisationId"])`

SamSalvatico (approved):

apps/messaging-api/src/routes/messages/index.ts
Thurfjell: yes, of course. I left it because I wasn't entirely sure. 

How grained should we go, should we have permissions for each field? 
eg. `messaging:message.partial.organisationId:read`
Pros
 - max flexibility
Cons 
 - lots of permissions

apps/messaging-api/src/services/messages/message-service.ts
Thurfjell: How will you know which fields then? Is there any particular reason you want to lock in, instead of keeping it flexible, but constrained? We have to create another getSlightlyDifferentPartialMessage function later (tm)? 
I think we should rename this function to getMessageOnlyOrganisationIdAndRecipientUserId in that case 😅

apps/messaging-api/src/routes/messages/index.ts
SamSalvatico: mmm I think just messaging:message.partial would be enough right now

apps/messaging-api/src/services/messages/message-service.ts
SamSalvatico: getPartialMessage is fine, we know that it just returns those fields, at the moment we don't have other needs, let's think on when we will need additional fields when we will need them

matteolc (approved):

PR #1532, opened 2025-04-03
27153 Dashboard admin update auth package to 4.0.4

Thurfjell:
### Ticket:

- [27153](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/27153)

### Description

Update the auth utils to match the 4.x version of our next auth pacakge. 

## Type

- [x] **Dependency upgrade**
- [ ] **Bug fix**
- [ ] **New feature**
- [x] **Dev change**
- [ ] **Additional tests**
- [ ] **Documentation**
- [ ] **Other**

### Checklist:

### Screenshots:

N/A


apps/dashboard-admin/components/UserContext.tsx
```diff
@@ -1,34 +1,23 @@
...
-    Pick<
-      AuthSessionContext,
-      "organization" | "isPublicServant" | "isInactivePublicServant"
-    >
+  user: AppUser
+  // AuthSessionUserInfo &
+  // Pick<
+  //   AuthSessionContext,
+  //   "organization" | "isPublicServant" | "isInactivePublicServant"
+  // >
```
Thurfjell, line 20:
```suggestion
```

apps/dashboard-admin/utils/logto-config.ts
```diff
@@ -86,8 +105,10 @@ const getConfig = () => {
       "messaging:event:read",
       "dashboard:admin:read",
       "dashboard:admin:write",
+      "profile:user:read",
```
SamSalvatico, line 108: We have to think a moment about this, because a ps will have this permission only if it has the profile public servant role, damn

Thurfjell: Yeah that's why I was thinking that the `profile:user.self:read` would be the universal "just let me get my own profile data" scope. 

Not sure this will necessarily interfere anything right now with sensitive profile PS role?

SamSalvatico: the problem is that we don't have that permission for organization roles (yet), but I think the only way is to add that there and set it for all the organization roles we have

Thurfjell: Agreed.

SamSalvatico (approved):

PR #1534, opened 2025-04-03
chore(AB#25422): Send secure message

matteolc:
### Ticket:

- [25422](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/25422)

### Notes

- [k8s configuration](https://github.com/ogcio/messaging-api-k8s-apps/pull/18)

### Description

image

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [ ] **New feature**
- [ ] **Dev change**
- [ ] **Additional tests**
- [ ] **Documentation**
- [ ] **Other**

### Checklist:

### Screenshots:

N/A


apps/messaging/app/[locale]/admin/send-a-message/createMessages.ts
```diff
@@ -24,66 +21,75 @@ export async function createMessages({
...
     }
   }
 
   let created = 0
-  const errors = {}
-  const profileSdk = BBClients.getProfileClient()
-  for (const userId of userIds) {
-    const { data: user, error } = await profileSdk.getProfile(userId)
+  const errors: Record<string, string> = {}
+  const profiles = await BBClients.getProfileClient().selectProfiles(
```
Comment (SamSalvatico) [inline apps/messaging/app/[locale]/admin/send-a-message/createMessages.ts:31]: I like this, but would be better to chunk into 100 items arrays?
Just to avoid problems on the url length

apps/messaging/const/messaging.ts
SamSalvatico, line 1: ❤️

apps/messaging/app/[locale]/admin/send-a-message/createMessages.ts
Comment (matteolc) [reply apps/messaging/app/[locale]/admin/send-a-message/createMessages.ts:31]: What problems could there be with the URL length? (we're not in a browser environment)

Comment (SamSalvatico) [reply apps/messaging/app/[locale]/admin/send-a-message/createMessages.ts:31]: It's okay like this at the moment, will do some fastify testing on this topic to check if handles everything as expected

Comment (SamSalvatico) [reply apps/messaging/app/[locale]/admin/send-a-message/createMessages.ts:31]: I've tested with 1000 ids, we should be fine

apps/messaging/app/[locale]/admin/providers/page.tsx
```diff
@@ -12,7 +11,7 @@ export default async (props: {
 }) => {
   const t = await getTranslations("settings.Page")
   const provider = props.searchParams?.provider
-  const isEmail = provider === searchValueEmail || !provider
+  const isEmail = provider === "email" || !provider
```
Comment (Thurfjell) [inline apps/messaging/app/[locale]/admin/providers/page.tsx:14]: Hehe just use the email check 😄

Thurfjell (dismissed):

Thurfjell (approved):

PR #1571, opened 2025-04-16
27587 - messaging api external id support

Thurfjell:
### Ticket:

- [27587](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/27587)

### Description

Add optional externalId string to POST /messages body.
If set, and scheduleAt is less than now (utc ms), we insert and instantly deliver the message, bypassing scheduler.
Else schedule as a normal message.

Added new event type import when we instant deliver messages. Reflected in a Tag in the UI.

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [x] **New feature**
- [x] **Dev change**
- [ ] **Additional tests**
- [ ] **Documentation**
- [ ] **Other**

### Checklist:

### Screenshots:

N/A


apps/messaging-api/src/services/messages/message-service.ts
```diff
@@ -120,6 +120,13 @@ export async function processMessage(params: {
 }) {
   const messagesProcessor = new MessagesProcessor(params.pool, params.logger);
 
+  const scheduledAt = new Date(params.message.scheduleAt)
+  const now = Date.now()
```
SamSalvatico, line 124: We need to add an environment variable, that, if set, let us know when digital_postbox stopped sending messages, so we can compare against it
If not set, use Date.now()

apps/messaging-api/src/services/messages/messages-processor.ts
```diff
@@ -9,12 +9,56 @@ import { ProfileM2MSdkWrapper } from "../users/profile-m2m-sdk-wrapper.js";
...
+        senderProfile,
+        senderApplication,
+        senderUser,
+        client,
+        messagingEventLogger,
+      });
+      messageId = id
+
+      messagingEventLogger.log(MessagingEventType.importedMessage, { messageId, externalId: message.externalId })
+      await setMessageAsDelivered({ client, messageId, recipientUserId: message.recipientUserId })
```
SamSalvatico, line 47: we also need to set message as seen

apps/messaging-api/src/services/messages/message-service.ts
```diff
@@ -120,6 +120,13 @@ export async function processMessage(params: {
...
+
+  if (params.message.externalId && (now > scheduledAt.getTime())) {
+    return messagesProcessor.processImportMessage({ message: params.message, senderUser: params.sender })
+  }
+
   return messagesProcessor.processMessage({
```
SamSalvatico, line 130: I would add the import event also in processMessage, to let us know it is an external message

apps/messaging-api/src/services/messages/messages-processor.ts
```diff
@@ -9,12 +9,56 @@ import { ProfileM2MSdkWrapper } from "../users/profile-m2m-sdk-wrapper.js";
...
+        recipientProfile,
+        senderProfile,
+        senderApplication,
+        senderUser,
+        client,
+        messagingEventLogger,
+      });
+      messageId = id
+
+      messagingEventLogger.log(MessagingEventType.importedMessage, { messageId, externalId: message.externalId })
```
SamSalvatico, line 46: I forgot, we need to set the timestamps in the events and also the created at equal to the date_time we got from the imported message

apps/messaging-api/src/services/messages/messages-processor.ts
Thurfjell: No we'll call the put endpoint for that.

SamSalvatico: What if the put fails? We would have an inconsistent status and we would double the number of requests
I would invoke that method here, given we know we are using this only for this use case

Thurfjell: @relvao

apps/messaging-api/src/services/messages/message-service.ts
Thurfjell: Why? At this point it's a normal message, no? @relvao

SamSalvatico: It would help maintaining a trace  about the fact that they have been imported from an external service

apps/messaging-api/src/services/messages/messages-processor.ts
Thurfjell: Are you sure about this?
@relvao  ?

apps/messaging-api/src/services/messages/message-service.ts
Thurfjell: Can we get a confirm on that. Help for whom? Who requested it?

PR #1624, opened 2025-05-12
AB#28331 fix a messaging choking pool issue

Thurfjell:
### Ticket:

- [28331](https://dev.azure.com/OGCIO-Digital-Services/Digital%20Services%20Programme/_workitems/edit/28331)

### Description

Fix issue with pg connections not being returned (suspected behavior at least) and the connection queue piling up prevents anyone from accessing the db.

## Type

- [ ] **Dependency upgrade**
- [x] **Bug fix**
- [ ] **New feature**
- [ ] **Dev change**
- [ ] **Additional tests**
- [ ] **Documentation**
- [ ] **Other**

### Checklist:

### Screenshots:

N/A


apps/messaging-api/src/routes/jobs/index.ts
```diff
@@ -30,7 +31,7 @@ export default async function jobs(app: FastifyInstance) {
       } finally {
         await eventLogger.commit();
       }
-
+      console.log("JOB HANDLER - BYE BYE - %s", request.id)
```
Thurfjell, line 34:
```suggestion
```

apps/messaging-api/src/routes/jobs/index.ts
```diff
@@ -16,6 +16,7 @@ export default async function jobs(app: FastifyInstance) {
       schema: ExecuteJobReqSchema,
     },
     async function jobHandler(request, reply) {
+      console.log("JOB HANDLER - HELLO - %s", request.id)
```
Thurfjell, line 19:
```suggestion
```

apps/messaging-api/src/plugins/external/pg.ts
```diff
@@ -12,6 +13,7 @@ export const autoConfig = (fastify: FastifyInstance) => {
     user: fastify.config.POSTGRES_USER,
     password: fastify.config.POSTGRES_PASSWORD,
     database: fastify.config.POSTGRES_DB_NAME,
+    connectionTimeoutMillis: 5000,
```
Thurfjell, line 16: This might be a risk in both ways. 

It timeout how long a connection will wait. Default is 0, wait forever.

Thurfjell, line 16:
```suggestion
```

Thurfjell:
> Could we abstract away `withClient` and `withRollback` wrappers to normalise how we deal with the pool across all apps and have a single place to deal with how to connect/release and what not?

I'm not a fan of this at all though. My opinion: node-pg have good abstractions so that the implementor can clearly see and understand what is going on. If there's any issue what so ever in the custom abstractions, it's, in my limited experience, magnitudes more tricky to deal with. Not saying these abstractions are complicated or whatever, just generic assumption and thoughts. I don't think we should arbitrary add abstractions on already abstracted libraries, and expect it to be standard.

matteolc (approved):

PR #1662, opened 2025-05-19
Chore: avoid logging email error

SamSalvatico:
- Removed logging sensitive data when an error is thrown sending email
- added EMAIL_PROVIDER_SMTP_USE_SSL with default true. "False" is needed to use maildev locally


apps/messaging-api/src/plugins/external/env.ts
```diff
@@ -33,11 +34,6 @@ export interface EnvConfig extends EnvDbConfig, EnvEmailConfig {
   LOGTO_M2M_UPLOADER_APP_SECRET: string;
   LOG_LEVEL: string;
   MESSAGING_SECURE_MESSAGE_URL: string;
-  EMAIL_PROVIDER_SMTP_HOST: string;
```
SamSalvatico, line 36: removed because are redundant, it already extends EnvEmailConfig

Thurfjell (approved):

PR #1663, opened 2025-05-19
fix: adding new screenshot

mikeclarkson:
### Ticket:

### Description

adds a new updated shot and moved the logout test i suspect is causing the footers link test to be flaky

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [ ] **New feature**
- [ ] **Dev change**
- [ ] **Additional tests**
- [ ] **Documentation**
- [ ] **Other**

### Checklist:

### Screenshots:

N/A


David-Losinski (approved):

apps/messaging/e2e/user/user.spec.ts
```diff
@@ -33,4 +29,8 @@ test.describe("User Features", () => {
       page.getByRole("heading", { name: "Teachtaireacht" }),
     ).toBeVisible()
   })
+
+  test("a user can logout @smoke @regression", async () => {
```
mikeclarkson, line 33: moved the logout test to the end of the file as i suspect it might have something to do with why the footer test was failing (it was always just stuck on the log in page and ran directly after this test)

PR #1770, opened 2025-07-01
chore(): fix create index concurrently migrations

matteolc:
### Ticket:

## Type

- [ ] **Dependency upgrade**
- [ ] **Bug fix**
- [ ] **New feature**
- [ ] **Dev change**
- [ ] **Additional tests**
- [ ] **Documentation**
- [ ] **Other**

### Checklist:

### Screenshots:

N/A


apps/messaging-api/src/migrations/sql/010.undo.messaging-create-trgm-ext.sql
```diff
@@ -0,0 +1,2 @@
+-- Undo GIN index on subject
```
matteolc, line 1: Fix this

PR #1781, opened 2025-07-10
Fix: manage missing region

SamSalvatico:


apps/messaging-api/src/services/jobs/job-service.ts
```diff
@@ -114,10 +114,6 @@ async function setJobAsRunning(params: {
     const jobResult = jobStatusResult.rows.at(0);
 
     if (!jobResult) {
-      eventLogger.log(
```
SamSalvatico, line 117: removed because messageId was missing, then was causing error on inserting on db

apps/messaging-api/src/services/jobs/job-service.ts
```diff
@@ -150,7 +146,7 @@ async function setJobAsRunning(params: {
     }
   } catch (err) {
     eventLogger.log(MessagingEventType.deliverMessageError, {
-      messageId: "",
```
SamSalvatico, line 153: if message id is set as empty an error is thrown because of not valid uuid when committing on db

apps/messaging-api/src/services/jobs/job-service.ts
```diff
@@ -163,7 +159,7 @@ async function setJobAsRunning(params: {
 
   if (!runningJob.userId || !runningJob.type) {
     eventLogger.log(MessagingEventType.deliverMessageError, {
-      messageId: runningJob?.jobId || "",
```
SamSalvatico, line 166: if message id is set as empty an error is thrown because of not valid uuid when committing on db

matteolc (approved):

